PR #17, opened 2014-07-24
break rerun if all protocols are passed or any of them is skipped

sergeychipiga:


florianpilz:
Was fixed by @davehunt. Before version 1.0.0 it was checked whether the test passed, now it is tested whether the test failed. Also, the test suite now makes sure skip and xfail work (both as marker and inside tests).

Issue #21, opened 2015-06-01
Python 3 compatibility


florianpilz:
Make the plugin work with Python 3.

florianpilz:
Did already work with Python 3, but tests were broken and horribly out of date. Thanks to @davehunt for rewriting all the tests. I really like the use of the new API which only talks about `outcomes`.

Issue #13, opened 2013-10-28
pytest-rerunfailures seeing skips as failures


bobsilverberg:
When running a test that includes a `@pytest.mark.skipif()`, when the test is skipped by pytest the plugin sees this as a failure and tries to rerun the test which is, of course, skipped again. With our Selenium tests, this causes a browser to open and then immediately close, which is unnecessary overhead (as is really any activity required to run and skip the test again).

The plugin should be updated to treat a skip as a pass so it will not be rerun.

florianpilz:
Hi bobsilverberg! We (gocept) have accepted the ownership of pytest-rerunfailures and are the new maintainers.

The problem you are encountering should be fixed, I agree. But it looks to me as if it is the same problem as stated by Pull Request https://github.com/gocept/pytest-rerunfailures/pull/17. There is already some work going on, so feel free to finish the tests so the PR can be merged.

florianpilz:
Was fixed by @davehunt. Before version 1.0.0 it was checked whether the test `passed`, not it is tested whether the test `failed`.

PR #23, opened 2015-11-24
Several fixes and enhancements

davehunt:
Lots of changes here, sorry for the big patch! I've addressed #13 and #21. It also makes #17 obsolete. Also a few new tests, and documentation!


davehunt:
It looks like there's an issue with pytest-xdist. I'll see if I can work out a fix.

davehunt:
I've added a workaround for the pytest-xdist issue, so I think this is good to go.

davehunt:
> Hi @davehunt, today is the day! I have big troubles to review your changes, since you mixed renames with changes. Therefore I will repeat your changes on a branch, but separate the rename part into a separate commit (and maybe split some more changes into commits). Please use smaller commits in the future. ;)

Sorry about that, one improvement led to another. I'm often on the other side, so appreciate that it can be difficult to review. Let me know if you need any help.

florianpilz:
@davehunt Why did you choose the Mozilla Public License?

florianpilz:
@klrmn Is it OK for you to put this project under the MIT license, which is used by `pytest` itself? Your are the author of the code, so it's your Copyright we are talking about.

florianpilz:
@davehunt @klrmn Ah, found the license declaration in `setup.py`. That clears things up.

florianpilz:
@davehunt I just released a new version on PyPI https://pypi.python.org/pypi/pytest-rerunfailures/1.0.0

davehunt:
Excellent, thank you! :smile:

klrmn:
I don't really have a strong opinion about which license (altho I probably
should). I just chose MPL because I was working with Mozilla at the time.

Issue #28, opened 2016-02-05
Not compatible with resultslog


Kanguros:
Hi,

The plugin is not compatible with the resultslog module from pytest.

By default the resultslog will not be created. To turn on this feature you need to set the `--resultlog` or `--result-log` pytest command line parameter. If we have a failing test and run pytest like so:

`$ pytest --result-log=pytest.log --reruns=3`

we will get the following error:

`...`
`INTERNALERROR>     res = hook_impl.function(*args)`
`INTERNALERROR>   File "c:\users\eostr\envs\pytest\lib\site-packages\_pytest\resultlog.py", line 86, in pytest_runtest_logreport`
`INTERNALERROR>     self.log_outcome(report, code, longrepr)`
`INTERNALERROR> UnboundLocalError: local variable 'longrepr' referenced before assignment`

This problem is cause because when a test fails for the first time (and not last time) line 86 is executed. The value of `report` will be set to `rerun`. In `pytest.pytest_runtest_logreport` there are a bunch of check that set the value of `longrepr` based on `report`. Since `rerun` is not valid value, `longrepr` does not get set.

Cheers,
Edgar

florianpilz:
Can you provide a test or even better a Pull Request to fix this issue? I will have no time for this issue during the following weeks.

klrmn:
or a PR to the README to document the incompatibility. when I wrote it, I
never expected it to be able to work with *every* pytest plugin, i just
knew that xdist was really important to work with.

PR #29, opened 2016-02-06
Fix for issue #28

Kanguros:
I created a fix for issue #28. Also added test for this case. Hope you like my changes.

### DIFF
--- a/pytest_rerunfailures.py
+++ b/pytest_rerunfailures.py
@@ -1,6 +1,7 @@
 import pytest
 
 from _pytest.runner import runtestprotocol
+from _pytest.resultlog import ResultLog
 
 
 # command line options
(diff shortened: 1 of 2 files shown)


florianpilz:
Thanks! I will have a look at your changes next week.

florianpilz:
OK, I see that this is the only viable option to solve the issue, since `ResultLog` cannot be teached new outcomes at the moment. But I think the problem should be fixed within `pytest` itself. Therefore I made a ticket in the `pytest` project: https://github.com/pytest-dev/pytest/issues/1419

Further steps depend on the outcome of that ticket.

florianpilz:
Since `ResultLog` will not be improved, I will merge your PR. Thanks for the work!

PR #40, opened 2016-09-26
Configure package as py2, py3 universal wheel

tltx:


sallner:
@tltx Thanks for the PR.

@florianpilz is leaving gocept and I will take over the maintenance for this package. So we are exercising all the permission and rights switches and a new release should be ready quite soon.

Issue #42, opened 2016-11-16
Add wait time


mblaschke:
For testing with infratest it would be nice to add a delay for reruns (in seconds)

icemac:
Nice idea. Pull requests are welcome.

Kanguros:
This look like something that I could give a go at. @icemac can you please assign this to me?

icemac:
@Kanguros Sorry for the long delay but I am not able to assign this issue to you: Your name does not show up in the Assignees list. 😱 I am not an administrator of the project so I am not able to look into the configuration. 
@sallner Are you able to assign this issue to @Kanguros?

sallner:
@icemac @Kanguros it seems that I can only assign members of the pytest organization to an issue. Have you gotten any further with the actual problem? Then we can have a look at it.

sallner:
@Kanguros What is the status of your contribution? Do you need help with anything?

Kanguros:
@sallner I finally found some time to work on this. I'll try to get this finished in a couple of days. I already have the command line argument working. Now going to try to add to marker.

sallner:
@Kanguros Thank you again for working on this feature. 

@mblaschke If you are still interested, I just released a 4.0 including this feature on pypi.

PR #55, opened 2017-12-10
Adding delay time between reruns, PR for #42

Kanguros:
This PR add the ability to specify a delay time between test reruns. The time can be specified as a command line argument or a with the "flaky" marker. 

Additionally some small maintenance work was. Tests and CI updated to support pytest 3.3.x

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -42,8 +42,8 @@ maximum number of times you'd like the tests to run:
   $ pytest --reruns 5
 
 To add a delay time between re-runs use the ``--reruns-delay`` command line
-option with amount of seconds that you would like wait before the next test
-re-run is lunched:
+option with the amount of seconds that you would like wait before the next
+test re-run is launched:
 
 .. code-block:: bash
 
@@ -66,7 +66,7 @@ test to run:
 Note that when teardown fails, two reports are generated for the case, one for
 the test case and the other for the teardown error.
 
-You can also add a delay time in the marker:
+You can also specify the re-run delay time in the marker:
 
 .. code-block:: python
 
--- a/README.rst
+++ b/README.rst
@@ -20,7 +20,7 @@ Requirements
 You will need the following prerequisites in order to use pytest-rerunfailures:
 
 - Python 2.7, 3.4, 3.5, 3.6, PyPy, or PyPy3
-- py.test 2.7.3 or newer
+- pytest 2.7.3 or newer
 
 Installation
 ------------
@@ -39,7 +39,15 @@ maximum number of times you'd like the tests to run:
 
 .. code-block:: bash
 
-  $ py.test --rerun 5
+  $ pytest --reruns 5
+
+To add a delay time between re-runs use the ``--reruns-delay`` command line
+option with amount of seconds that you would like wait before the next test
+re-run is lunched:
+
+.. code-block:: bash
+
+   $ pytest --reruns 5 --reruns-delay 1
 
 Re-run individual failures
 --------------------------
@@ -57,6 +65,15 @@ test to run:
 
 Note that when teardown fails, two reports are generated for the case, one for
 the test case and the other for the teardown error.
+
+You can also add a delay time in the marker:
+
+.. code-block:: python
+
+  @pytest.mark.flaky(reruns=5, reruns_delay=2)
+  def test_example():
+      import random
+      assert random.choice([True, False])
 
 Output
 ------
--- a/pytest_rerunfailures.py
+++ b/pytest_rerunfailures.py
@@ -68,12 +68,10 @@ def check_options(config):
         config.pluginmanager.register(config._resultlog)
 
 
-def pytest_runtest_protocol(item, nextitem):
-    """
-    Note: when teardown fails, two reports are generated for the case, one for
-    the test case and the other for the teardown error.
-    """
+def reruns_count(item):
     rerun_marker = item.get_marker("flaky")
+    reruns = None
+
     # use the marker as a priority over the global setting.
     if rerun_marker is not None:
         if "reruns" in rerun_marker.kwargs:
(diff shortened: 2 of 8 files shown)


README.rst
```diff
@@ -39,7 +39,15 @@ maximum number of times you'd like the tests to run:
 
 .. code-block:: bash
 
-  $ py.test --rerun 5
+  $ pytest --reruns 5
+
+To add a delay time between re-runs use the ``--reruns-delay`` command line
+option with amount of seconds that you would like wait before the next test
+re-run is lunched:
```
tomviner: :hamburger: :smile: 

Pretty sure you mean _launched_!

Kanguros: :) Probably I was very hungry at the time when writing that. How should I go about fixing that, new PR?

tomviner: Haha. No need for a new PR, just push another commit to your branch and it'll show up here.

test_pytest_rerunfailures.py
```diff
@@ -219,3 +221,32 @@ def test_fail():
...
+        def test_fail_one():
+            assert False
+            
+        @pytest.mark.flaky(reruns=2, reruns_delay=1)
+        def test_fail_two():
+            assert False""")
+
+    result = testdir.runpytest()
+
+    assert_outcomes(result, passed=0, failed=2, rerun=3)
```
tomviner: So these tests are good to check nothing blows up and the tests are run the correct number of times, but do we need a check that the delay is actually getting put in place?

Did you consider we could mock out time.sleep, and assert it's called with the correct value? This would have the side benefit of speeding up the test run - not that that really matters here :-)

Kanguros: Yes, that a good idea. For this should I just use `unittest.mock` or is there a way to do it with the `monkeypatch` fixture?

README.rst
```diff
@@ -39,7 +39,15 @@ maximum number of times you'd like the tests to run:
 
 .. code-block:: bash
 
-  $ py.test --rerun 5
+  $ pytest --reruns 5
```
sallner: Thank you for fixing the documentation here.

pytest_rerunfailures.py
```diff
@@ -59,12 +68,10 @@ def check_options(config):
         config.pluginmanager.register(config._resultlog)
 
 
-def pytest_runtest_protocol(item, nextitem):
-    """
-    Note: when teardown fails, two reports are generated for the case, one for
-    the test case and the other for the teardown error.
-    """
+def reruns_count(item):
```
sallner: I generally prefer to have function names which start with a verb, e.g. `get_maximum_reruns` to enhance readability of the code.

pytest_rerunfailures.py
```diff
@@ -78,15 +85,43 @@ def pytest_runtest_protocol(item, nextitem):
     elif item.session.config.option.reruns:
         # default to the global setting
         reruns = item.session.config.option.reruns
+
+    return reruns
+
+
+def reruns_delay(item):
```
sallner: I would also prefer a verb at the start of this function name.

test_pytest_rerunfailures.py
```diff
@@ -219,3 +221,32 @@ def test_fail():
                                '--result-log', './pytest.log')
 
     assert_outcomes(result, passed=0, failed=1, rerun=2)
+
+
+@pytest.mark.parametrize('delay_time', [0, 0.0, 1, 2.5])
```
sallner: I like the use of  the parametrize very much. As this is now easily possible I would like to test here also the unhappy path of a negative input (-1). I presume that this will fail during `time.sleep()`.

I know, that there is currently also no plausibility check for negative inputs on `reruns` but this is a topic for a different issue, as it is not failing but silently not collecting any tests.

test_pytest_rerunfailures.py
sallner: That sounds like a good idea, +1.

pytest_rerunfailures.py
```diff
@@ -34,6 +35,14 @@ def pytest_addoption(parser):
...
         default=0,
         help="number of times to re-run failed tests. defaults to 0.")
+    group._addoption(
+        '--reruns-delay',
+        action='store',
+        dest='reruns_delay',
+        type=float,
+        default=0,
+        help='add time (seconds) delay between reruns.'
+    )
```
sallner: In addition to the addition to the command line it is also helpful to have `reruns_delay` added in the documentation for the config. https://github.com/pytest-dev/pytest-rerunfailures/pull/55/files#diff-c3a581865b7c0be4e6b5d5fa60d6aed5R51

test_pytest_rerunfailures.py
Kanguros: Not sure what you mean here. Should I some kind of `delay_time` validation to the plugin code, so that it will some how handle a negative value?

test_pytest_rerunfailures.py
Kanguros: For Python3 I will use `unittest.mock` but for Python2 this requires the `mock` package to be installed. Since this package is only required for the test I don't see any point in adding it to `setup.py`.  Is it ok if I just add it as a dependency in `tox.ini`?

Kanguros:
Any idea why the `mock` is not working with pytest==2.7.3?

test_pytest_rerunfailures.py
sallner: That's exactly what I meant. I would prefere to the code to work as if there would be no `delay_time` set, in case of a negative delay time and give a warning.

sallner:
@Kanguros It looks like `pytest=2.7.3` might spawn a subprocess for `runpytest()`, which in turn makes the mock obsolete. I would suggest we drop the support for it as it is already old enough. In addition to remove it from the `tox.ini`, could you please update the minimum version in the setup.py?

Kanguros:
@sallner Dropped support for pytest==2.7.3. All the tests are now passing.

sallner:
@Kanguros Thank you again for your work. I just released a 4.0 on pypi.

PR #59, opened 2018-05-21
Use item.get_closest_marker if possible

The-Compiler:
This avoids deprecation warnings on the upcoming pytest 3.6 release:
https://docs.pytest.org/en/features/mark.html#marker-revamp-and-iteration

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,8 +5,10 @@ Changelog
 4.1 (unreleased)
 ================
 
-- Nothing changed yet.
+- Add support for pytest 3.6 by using ``Node.get_closest_marker()`` (Thanks to
+  `@The-Compiler`_ for the PR).
 
+.. _@The-Compiler: https://github.com/The-Compiler
 
 4.0 (2017-12-23)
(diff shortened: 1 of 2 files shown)


sallner:
looks good to me, thanks for this proactive change. @The-Compiler could you please add a changelog entry with attribution and then we can merge it.

The-Compiler:
Oh, right - done!

sallner:
Fine, I think a release is not necessary until the final release of pytest 3.6.

The-Compiler:
pytest 3.6.0 [is out now](https://pypi.org/project/pytest/#history) - so I'd appreciate a release :smile:

sallner:
@The-Compiler The is also a fresh release of [pytest-rerunfailures](https://pypi.org/project/pytest-rerunfailures/).

The-Compiler:
The last release I see there is 4.0 from December.

The-Compiler:
Huh, nevermind - with a forced reload, I now see the new release... Thanks a lot!

PR #66, opened 2018-09-08
[FIX] Added cleaning steps for cashed fixture result and failed setup states.

### LINKED ISSUE #64: weird setup_class behavior with rerun
test_rerun.py

```
class Test1(object):

    def setup_class(cls):
        print 'setup_class start'
        raise Exception('setup class failure')

    def test1(self):
        pass

    def test2(self):
        pass

    def test3(self):
        pass

```

`pytest               2.9.1
pytest-rerunfailures 4.1`
`py.test --reruns=1 test_rerun.py -v -s`

> test_rerun.py::Test1::test1 setup_class start
> RERUN
> test_rerun.py::Test1::test1 ERROR
> test_rerun.py::Test1::test2 RERUN
> test_rerun.py::Test1::test2 ERROR
> test_rerun.py::Test1::test3 RERUN
> test_rerun.py::Test1::test3 setup_class start
> ERROR

The output looks confusing. It looks the setup_class rerun happens only for test3.

OlegKuzovkov:
fixes #64  issue for:
- setup_class method
- session\module\class scope fixtures

### DIFF
--- a/pytest_rerunfailures.py
+++ b/pytest_rerunfailures.py
@@ -146,7 +146,7 @@ def _remove_failed_setup_state_from_session(item):
     for col in setup_state.stack:
         if hasattr(col, prepare_exc):
             delattr(col, prepare_exc)
-    setup_state.stack.clear()
+    setup_state.stack = list()
 
 
 def pytest_runtest_protocol(item, nextitem):
--- a/pytest_rerunfailures.py
+++ b/pytest_rerunfailures.py
@@ -122,7 +122,7 @@ def get_reruns_delay(item):
     return delay
 
 
-def _remove_cached_results_from_fixtures(item):
+def _remove_cached_results_from_failed_fixtures(item):
     """
     Note: remove all cached_result attribute from every fixture
     """
@@ -132,7 +132,9 @@ def _remove_cached_results_from_fixtures(item):
         fixture_defs = fixture_info.name2fixturedefs[fixture_def_str]
         for fixture_def in fixture_defs:
             if hasattr(fixture_def, cached_result):
-                delattr(fixture_def, cached_result)
+                result, cache_key, err = getattr(fixture_def, cached_result)
+                if err:  # Deleting cached results for only failed fixtures
+                    delattr(fixture_def, cached_result)
 
 
 def _remove_failed_setup_state_from_session(item):
@@ -144,7 +146,7 @@ def _remove_failed_setup_state_from_session(item):
     for col in setup_state.stack:
         if hasattr(col, prepare_exc):
             delattr(col, prepare_exc)
-    setup_state.stack = list()
+    setup_state.stack.clear()
 
 
 def pytest_runtest_protocol(item, nextitem):
@@ -186,7 +188,7 @@ def pytest_runtest_protocol(item, nextitem):
                     item.ihook.pytest_runtest_logreport(report=report)
 
                 # cleanin item's cashed results from any level of setups
-                _remove_cached_results_from_fixtures(item)
+                _remove_cached_results_from_failed_fixtures(item)
                 _remove_failed_setup_state_from_session(item)
 
                 break  # trigger rerun
--- a/pytest_rerunfailures.py
+++ b/pytest_rerunfailures.py
@@ -122,6 +122,30 @@ def get_reruns_delay(item):
     return delay
 
 
+def _remove_cached_results_from_fixtures(item):
+    """
+    Note: remove all cached_result attribute from every fixture
+    """
+    cached_result = 'cached_result'
+    fixture_info = getattr(item, '_fixtureinfo')
+    for fixture_def_str in fixture_info.name2fixturedefs:
+        fixture_defs = fixture_info.name2fixturedefs[fixture_def_str]
+        for fixture_def in fixture_defs:
+            if hasattr(fixture_def, cached_result):
+                delattr(fixture_def, cached_result)
+
+
+def _remove_failed_setup_state_from_session(item):
+    """
+    Note: remove all _prepare_exc attribute from every col in stack of _setupstate
+    """
+    prepare_exc = "_prepare_exc"
+    setup_state = getattr(item.session, '_setupstate')
+    for col in setup_state.stack:
+        if hasattr(col, prepare_exc):
+            delattr(col, prepare_exc)
+
+
 def pytest_runtest_protocol(item, nextitem):
     """
     Note: when teardown fails, two reports are generated for the case, one for
@@ -159,6 +183,10 @@ def pytest_runtest_protocol(item, nextitem):
                 if not parallel or works_with_current_xdist():
                     # will rerun test, log intermediate result
                     item.ihook.pytest_runtest_logreport(report=report)
+
+                # cleanin item's cashed results from any level of setups
+                _remove_cached_results_from_fixtures(item)
+                _remove_failed_setup_state_from_session(item)
 
                 break  # trigger rerun
         else:
--- a/pytest_rerunfailures.py
+++ b/pytest_rerunfailures.py
@@ -137,13 +137,14 @@ def _remove_cached_results_from_fixtures(item):
 
 def _remove_failed_setup_state_from_session(item):
     """
-    Note: remove all _prepare_exc attribute from every col in stack of _setupstate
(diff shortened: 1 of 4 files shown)


pytest_rerunfailures.py
```diff
@@ -122,6 +122,31 @@ def get_reruns_delay(item):
...
+    cached_result = 'cached_result'
+    fixture_info = getattr(item, '_fixtureinfo')
+    for fixture_def_str in fixture_info.name2fixturedefs:
+        fixture_defs = fixture_info.name2fixturedefs[fixture_def_str]
+        for fixture_def in fixture_defs:
+            if hasattr(fixture_def, cached_result):
+                delattr(fixture_def, cached_result)
+
+
+def _remove_failed_setup_state_from_session(item):
```
sallner: I am a little bit reluctant to introduce cleanup login in the plugin. This could lead to too much coupling. As the `SetupState` is not too complicated, it looks like https://github.com/pytest-dev/pytest/blob/master/src/_pytest/runner.py#L331 is actually doing the job already. Would that be a possibility?

pytest_rerunfailures.py
```diff
@@ -160,6 +185,10 @@ def pytest_runtest_protocol(item, nextitem):
                     # will rerun test, log intermediate result
                     item.ihook.pytest_runtest_logreport(report=report)
 
+                # cleanin item's cashed results from any level of setups
+                _remove_cached_results_from_fixtures(item)
+                _remove_failed_setup_state_from_session(item)
```
sallner: What I am not sure about is the effect for the result of a module or session scoped fixture. Could you please add a test for them as well, that they are not affected by this change? I guess this should not re-initiate those fixtures as they could be expensive database fixtures.

test_pytest_rerunfailures.py
```diff
@@ -267,3 +267,42 @@ def test_fail_two():
...
+                    assert False
+                assert True
+            @pytest.mark.parametrize('param', [1, 2, 3])
+            def test_pass(self, param):
+                assert param""")
+    result = testdir.runpytest('--reruns', '1')
+    assert_outcomes(result, passed=3, rerun=1)
+
+
+def test_rerun_on_class_scope_fixture_setup_with_first_execution_error_for_parametrized_test(testdir):
```
sallner: The function names of these tests are quite long. Could you please shorten them and write the more verbose description in a docstring? This reminds me to add a flake8 test to the tox run.

pytest_rerunfailures.py
OlegKuzovkov: If I understand correctly, you advising to use SetupState.teardown_all() method, instead of a custom cleaner that I've implemented.

I've tried to use it, but it seems not handling the resource properly. All it does is calling all the `teadowns` for each `col` elements in the stack, but it is missing the most important part: invalidation/removing the failed execution result stored in `_prepare_exc` field. That's why every new test that is relying on `setup_class` will rerun it anyway, but it shouldn't because setup was already successfully re-executed after the first attempt.
You can see the problem by changing the body of `_remove_failed_setup_state_from_session` function to: `setup_state = getattr(item.session, '_setupstate') setup_state.teardown_all()` and running the test:
'''
class TestFoo(object):
    execution_number = 1
    srr = None

    @classmethod
    def setup_class(cls):
        cls.srr = Random.rand_char_string(10)
        Log.info('IN SETUP %s' % cls.srr)
        if cls.execution_number:
            cls.execution_number -= 1
            assert False
        assert True

    @mark.parametrize('param', [1, 2, 3])
    def test_pass(self, param):
        Log.info('IN TEST %s %s' % (param, self.srr))
        assert param
'''

The output will be:
'''
R2018-09-14 11:12:27,314 IN SETUP XWnfETJrmi
2018-09-14 11:12:27,568 IN SETUP ArdcndHZkT
.2018-09-14 11:12:27,569 IN TEST 1 ArdcndHZkT
R2018-09-14 11:12:27,617 IN SETUP GbnwDwvSWi
.2018-09-14 11:12:27,618 IN TEST 2 GbnwDwvSWi
R2018-09-14 11:12:27,642 IN SETUP tdWVHVPPyi
.2018-09-14 11:12:27,643 IN TEST 3 tdWVHVPPyi
'''

I do agree with you though, that instead of cleaning stack by myself, I should've used teardown_all().
With this change the output of the previous test is correct:
'''
R2018-09-14 11:13:36,147 IN SETUP biCSDlQwSP
2018-09-14 11:13:36,303 IN SETUP ZKlZjHlYDB
.2018-09-14 11:13:36,304 IN TEST 1 ZKlZjHlYDB
.2018-09-14 11:13:36,308 IN TEST 2 ZKlZjHlYDB
.2018-09-14 11:13:36,320 IN TEST 3 ZKlZjHlYDB
'''

test_pytest_rerunfailures.py
OlegKuzovkov: I will be really appreciated if you could suggest me the names for these cases. I really cannot find any clearer and shorter test names for this specific contexts :)

pytest_rerunfailures.py
OlegKuzovkov: Further investigation showed that we cannot use `.teardown_all()` because it is triggering all the finalizers, even for the module/session scoped fixtures. Thus, leaving original implementation.

pytest_rerunfailures.py
```diff
@@ -132,7 +132,9 @@ def _remove_cached_results_from_fixtures(item):
         fixture_defs = fixture_info.name2fixturedefs[fixture_def_str]
         for fixture_def in fixture_defs:
             if hasattr(fixture_def, cached_result):
-                delattr(fixture_def, cached_result)
+                result, cache_key, err = getattr(fixture_def, cached_result)
+                if err:  # Deleting cached results for only failed fixtures
+                    delattr(fixture_def, cached_result)
```
sallner: I like the test for an error. Hopefully we will not have a chain of dependent fixtures and strange results, but then the dependent fixture should also produce an error.

pytest_rerunfailures.py
sallner: Yeah, that is true. So I am happy with keeping the cleanup here. At this point a comment could be helpful.

pytest_rerunfailures.py
sallner: I would not like to re-initialize all fixtures, but only the one who failed. So that should be the case with the current state of the PR.

test_pytest_rerunfailures.py
```diff
@@ -288,7 +291,10 @@ def test_pass(self, param):
     assert_outcomes(result, passed=3, rerun=1)
 
 
-def test_rerun_on_class_scope_fixture_setup_with_first_execution_error_for_parametrized_test(testdir):
+def test_rerun_on_class_scope_fixture_with_error_with_reruns(testdir):
+    """
+    Case: Class scope fixture throwing error on the first execution for parametrized test
+    """
```
sallner: Thank you for rephrasing the function names.

test_pytest_rerunfailures.py
```diff
@@ -306,3 +312,65 @@ def test_pass(self, setup_fixture, param):
...
+def test_rerun_on_module_fixture_with_reruns(testdir):
+    """
+    Case: Module scope fixture is not re-executed when class scope fixture throwing error on the first execution
+    for parametrized test
+    """
+    testdir.makepyfile("""
+        import pytest
+        
+        class Execution:
+            count = 0
```
sallner: I like those two tests and I see that you have this pattern of execution number, but I seems more pythonic to me to use a global boolean in this case instead of a singleton class to store it. But that is on you.

pytest_rerunfailures.py
OlegKuzovkov: I've added the test for the error (test_rerun_on_module_fixture_with_reruns). Only setup_fixture in that scenario will fail and will be cleaned up. Dependent fixture cannot produce any error if one of the fixtures before fails.
Unless I am missing something.

pytest_rerunfailures.py
OlegKuzovkov: That is correct. The only failed fixture will be erased from any previous results.

test_pytest_rerunfailures.py
OlegKuzovkov: Thank you for the suggestion. Will redo this.

PR #68, opened 2018-09-13
[FEAT] Added execution_count variable.

### LINKED ISSUE #67: Expose execution count on the item level.
It would be great to know the number of executions of a current item (test case) stored in the internal variable of the object.

We have built an integration with a Test Cases Management system in our company, and for the flaky tests that were passed after a certain retry, we would like to mark them with a different status rather than always "passed".

OlegKuzovkov:
fixes #67 :
 - Added execution_count variable to the "item" object level
 - Changed a general loop logic according to pythonic way of do/while implementation.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,7 +5,10 @@ Changelog
 4.2 (unreleased)
 ================
 
-- Nothing changed yet.
+- Added new ``execution_count`` attribute to reflect the number of test case executions according to #67 issue.
+  (Thanks to `@OlegKuzovkov`_ for the PR).
+
+.. _@OlegKuzovkov: https://github.com/OlegKuzovkov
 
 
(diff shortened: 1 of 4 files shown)


sallner:
Hi @OlegKuzovkov , thanks for the PR and the improvement. It looks like a good idea to work with. To before merging this PR, I would like you to add a changelog entry. If you have time, it would also be great, if you could add a small paragraph in the README.rst, maybe under a new section `Development.`

sallner:
Thank you for the changes. I like that you put an example in the README.

darshakkakkad:
@OlegKuzovkov @sallner do you know when would these changes be released?

sallner:
@darshakkakkad @OlegKuzovkov I just put out a new release 4.2 for `pytest-rerunfailures`, you can use it now.

PR #70, opened 2018-10-17
Test with current pytest and Python version.

sallner:
Drop legacy versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ python:
   - 3.6
   - 3.7
   - 3.7-dev
-  - pypy2.7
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -4,6 +4,7 @@ python:
   - 3.4
   - 3.5
   - 3.6
+  - 3.7
```
icemac: 3.7 needs `xenial` and `sudo`, see https://github.com/zopefoundation/zope.annotation/pull/17/files#diff-354f30a63fb0907d4ad57269548329e3R9 as an example

Issue #77, opened 2019-01-05
Lack of rerun attribute on report since 4.2 version


RibeiroAna:
Some Pytest-HTML tests were failing since October and I realised that it was mainly because of that specific line: 

`test_index = hasattr(report, 'rerun') and report.rerun + 1 or 0`

The problem is that until version 4.1, the report had a rerun attribute, but for some reason it was taken off in version 4.2.  I wish that it could come back.

RibeiroAna:
@sallner before closing the issue, just one question: when do you intend to release version 6?

sallner:
@RibeiroAna I will add some maintenance stuff and should cut a release later today.

sallner:
The new release including a fix for this regression can be found here https://pypi.org/project/pytest-rerunfailures/6.0/

PR #78, opened 2019-01-05
Fixes issue related to rerun attribute on report (Issue #77)

RibeiroAna:
As I reported on #77


pytest_rerunfailures.py
```diff
@@ -175,6 +175,7 @@ def pytest_runtest_protocol(item, nextitem):
         reports = runtestprotocol(item, nextitem=nextitem, log=False)
 
         for report in reports:  # 3 reports: setup, call, teardown
+            report.rerun = item.execution_count
```
sallner: To restore the old behaviour, I think we need to subtract from the `execution_count`.
```suggestion
            report.rerun = item.execution_count -1
```

test_pytest_rerunfailures.py
```diff
@@ -388,3 +388,13 @@ def pytest_runtest_teardown(item):
             assert item.execution_count == 3""")
     result = testdir.runpytest('--reruns', '2')
     assert_outcomes(result, passed=3, rerun=2)
+
+def test_rerun_report(testdir):
+    testdir.makepyfile('def test_pass(): assert False')
+    testdir.makeconftest("""
+        def pytest_runtest_logreport(report):
+            assert hasattr(report, 'rerun')
+            assert isinstance(report.rerun, int)
```
sallner: Maybe we could assert the upper limit for reruns here. 
```suggestion
            assert isinstance(report.rerun, int)
            assert report.rerun <= 2
```

sallner:
Thanks again for you contribution.

PR #98, opened 2019-11-13
Explicitly mark dependency on `setuptools`

Eric-Arellano:
`pytest-rerunfailures` explicitly imports `pkg_resources`, which is only provided via `setuptools`:

https://github.com/pytest-dev/pytest-rerunfailures/blob/e3eae032cc75104649a1ba17b22d82835ab8bda9/pytest_rerunfailures.py#L1

While most Python distributions come bundled with this distribution, not all do and it is more robust to make explicit this dependency in `setup.py`. Otherwise, if the distribution does not have `setuptools` already installed, pytest will fail at runtime because it cannot find `pkg_resources`.

The version is pinned at `>=40.0` to match the version used in `pytest`: https://github.com/pytest-dev/pytest/blob/55a58bcd3d15887610e1303f4076401bd16ba742/setup.py#L22

(P.S. Thanks for this plugin! It's exactly what I need for my project 🎉)


sallner:
@Eric-Arellano I created a new release of the package. https://pypi.org/project/pytest-rerunfailures/8.0/

Eric-Arellano:
Thank you @sallner! I appreciate you merging this and promptly releasing!

PR #110, opened 2020-06-03
Configure python_requires

nicoddemus:
This prevents pip from installing pytest-rerunfailures in unsupported Python versions

https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires


sallner:
@nicoddemus Thank you for the extra step!

PR #111, opened 2020-07-01
README: remove incorrect hookwrapper

bluetech:
A hookwrapper requires a `yield`, but seems unnecessary here, so just remove it.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -127,6 +127,6 @@ Development
 
   .. code-block:: python
 
-    @hookimpl(tryfirst=True, hookwrapper=True)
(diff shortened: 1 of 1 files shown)


icemac:
Thank you for this PR.
I have one question to understand the reason behind: What are the consequences for keeping this parameter?

bluetech:
There will be an error when this hook is called. If https://github.com/pytest-dev/pluggy/pull/282 is merged, then it will be an error when the plugin is registered.

icemac:
Thank you for the link and providing the PR beforehand. I am going to merge it now.

Issue #101, opened 2020-02-17
Rerun only for certain errors?


deeplook:
I've discovered this pytest plugin, which looks very interesting to me, but ideally I would want it to rerun tests a little more fine-grained, namely only when specific errors are detected, say HTTP response codes 5xx. I don't think this is already possible, right? If not maybe consider it as a feature request?

icemac:
It is not yet possible to rerun on specific exceptions. Maybe you would like to come up with a pull request.

gnikonorov:
Hey @icemac @deeplook what were we thinking? How about to start a flag that is a comma separated string of exception names ( e.g: `--only-rerun=ValueError,BufferError,MyCustomError` )

gnikonorov:
Hey @icemac I self assigned this to myself since I want to make a PR for this. My apologies if this was unwelcome

icemac:
@gnikonorov This is perfectly welcome. I was no planning to work on this issue. Thank you for tackling it.

Issue #43, opened 2016-11-29
Add optiontional "minimum number of passes" to consider a pass


lexrj:
Currently rerunning a test n times is allowed, but as long as it passes once, the test is considered passed. It would be helpful to have a minimum number of passes required for the test to be considered a pass, for more nuanced pass rates.

Example: I know something fails 33% of the time, so I want to run it four times and make sure it passes at least twice, otherwise I should look into it.

icemac:
This seems to be an interesting idea. A pull request is welcome.

kousu:
For inspiration, [flaky](https://github.com/box/flaky)'s API for this looks like:

```
@flaky(max_runs=3, min_passes=2)
```

gnikonorov:
Hey @icemac do you mind if I grab this one too?

icemac:
@gnikonorov Feel free to tackle this one, too.

icemac:
I am closing this old issue. Feel free to re-open it if there are any news about the topic.

PR #113, opened 2020-07-14
Add --only-rerun flag to add option to rerun specific errors

gnikonorov:
Add new flag `--only-rerun` to allow for users to only rerun errors matching regexes passed to this flag. If the flag is omitted or passed in without a value then no changes occur and all failures are rerun.

Closes https://github.com/pytest-dev/pytest-rerunfailures/issues/101

### DIFF
--- a/pytest_rerunfailures.py
+++ b/pytest_rerunfailures.py
@@ -33,11 +33,11 @@ def pytest_addoption(parser):
         "re-run failing tests to eliminate flaky failures")
     group._addoption(
         '--only-rerun',
-        action='store',
+        action='append',
         dest='only_rerun',
         type=str,
         default=None,
-        help='Optional comma separated list. If passed, only rerun errors matching the regexes provided.'
+        help='If passed, only rerun errors matching the regex provided. Pass this flag multiple times to accumulate a list of regexes to match'
     )
     group._addoption(
         '--reruns',
@@ -190,7 +190,7 @@ def _should_hard_fail_on_error(session_config, report):
     if not rerun_errors:
         return False
 
-    for rerun_regex in rerun_errors.split(','):
+    for rerun_regex in rerun_errors:
         if re.search(rerun_regex, report.longrepr.reprcrash.message):
             return False
 
--- a/pytest_rerunfailures.py
+++ b/pytest_rerunfailures.py
@@ -1,4 +1,5 @@
 import pkg_resources
+import re
 import time
 import warnings
 
@@ -30,6 +31,14 @@ def pytest_addoption(parser):
     group = parser.getgroup(
         "rerunfailures",
         "re-run failing tests to eliminate flaky failures")
+    group._addoption(
+        '--only-rerun',
+        action='store',
+        dest='only_rerun',
+        type=str,
+        default=None,
+        help='Optional comma seperated list. If passed, only rerun errors matching the regexes provided.'
+    )
     group._addoption(
         '--reruns',
         action="store",
@@ -173,6 +182,21 @@ def _remove_failed_setup_state_from_session(item):
     setup_state.stack = list()
 
 
+def _should_fail_on_error(session_config, report):
+    if report.outcome != 'failed':
+        return False
+
+    rerun_errors = session_config.option.only_rerun
+    if not rerun_errors:
+        return False
+
+    for rerun_regex in rerun_errors.split(','):
+        if re.search(rerun_regex, report.longrepr.reprcrash.message):
+            return False
+
+    return True
+
+
 def pytest_runtest_protocol(item, nextitem):
     """
     Note: when teardown fails, two reports are generated for the case, one for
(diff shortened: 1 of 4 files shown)


pytest_rerunfailures.py
```diff
@@ -30,6 +31,14 @@ def pytest_addoption(parser):
     group = parser.getgroup(
         "rerunfailures",
         "re-run failing tests to eliminate flaky failures")
+    group._addoption(
+        '--only-rerun',
+        action='store',
+        dest='only_rerun',
+        type=str,
+        default=None,
+        help='Optional comma separated list. If passed, only rerun errors matching the regexes provided.'
```
icemac, line 40: I was wondering if there is the ability to use this option multiple times, so we get a list of regexes from the parser. This would allow commas in the regex and the shell eliminates not necessary whitespace for us.

gnikonorov: Let me check tonight ( I’m in EST )

gnikonorov: setting action=append for `group._addoption` does that. Thank you for bringing this up, I updated the PR accordingly

gnikonorov:
I updated this PR according to your feedback @icemac

test_pytest_rerunfailures.py
```diff
@@ -409,27 +409,31 @@ def pytest_runtest_logfinish(nodeid, location):
...
+            ('def test_only_rerun(): raise AssertionError("ERR")', ['Assertion*'], True),
+            ('def test_only_rerun(): raise AssertionError("ERR")', ['Assertion'], True),
+            ('def test_only_rerun(): raise AssertionError("ERR")', ['ValueError'], False),
+            ('def test_only_rerun(): raise AssertionError("ERR")', [''], True),
+            ('def test_only_rerun(): raise AssertionError("ERR")', ['AssertionError: '], True),
+            ('def test_only_rerun(): raise AssertionError("ERR")', ['AssertionError: ERR'], True),
+            ('def test_only_rerun(): raise AssertionError("ERR")', ['ERR'], True),
+            ('def test_only_rerun(): raise AssertionError("ERR")', ['AssertionError,ValueError'], False),
+            ('def test_only_rerun(): raise AssertionError("ERR")', ['AssertionError ValueError'], False),
+            ('def test_only_rerun(): raise AssertionError("ERR")', ['AssertionError','ValueError'], True),
```
icemac, line 424:
```suggestion
            ('def test_only_rerun(): raise AssertionError("ERR")', ['AssertionError', 'ValueError'], True),
```

gnikonorov:
Thank you. I'll merge this, and open a a PR for the README and changelog.

There's no rush for a release, thank you for asking.

gnikonorov:
Added the README and changelog changes here, since it doesn't make sense to do otherwise. Thanks for reminding me @icemac. Please let me know if you're still comfortable with me merging these changes

pytest_rerunfailures.py
```diff
@@ -173,6 +182,21 @@ def _remove_failed_setup_state_from_session(item):
     setup_state.stack = list()
 
 
+def _should_hard_fail_on_error(session_config, report):
+    if report.outcome != 'failed':
```
sallner, line 186: It seems to me, that this test is also done in line 230 with `report.failed` or am I missing some edge case here? Which test case would fail, if we leave it out?

gnikonorov: This is unrelated since it says that anything that is not failed cannot be a hard failure. Which makes sense to me, since otherwise this function would potentially flag non failed things as hard failures

sallner: I understand the reason but I do not see, why we need it twice currently. It seems to be a guard for a case, that is not possible at the moment. But I would like to refactor line 230 anyhow, as this is not so easy to comprehend any more.

gnikonorov: I’m working on [this issue](https://github.com/pytest-dev/pytest-rerunfailures/issues/43) and in the change I have a refactor of that line.

This change came out of it. It’s just unrelated, so I raised it first.

graingert:
@S-Borkowski https://github.com/pytest-dev/pytest-rerunfailures/compare/9.0...master

S-Borkowski:
When I run it like this: 
```
pytest -n auto --dist loadscope --reruns 3 --only-rerun PreSubmitError pathToTest
```
I get and error: pytest: error: unrecognised arguments: --only-rerun pathToTest

S-Borkowski:
Downloaded version 9.0 (most recent) from pypi. I am unable to install packages through console; i need .whl file to install

The-Compiler:
@S-Borkowski 9.0 was released on March 18th, this PR was merged on July 17th. You'll need to either find a way to install from the git repository, or wait until the next release.

S-Borkowski:
@The-Compiler When will approx. be the next release? Is there any way You could provide me with .whl of July's version?

The-Compiler:
@S-Borkowski  Here's the result of running `python3 setup.py bdist_wheel` in this repo: 
[pytest_rerunfailures-9.1.dev0-py3-none-any.zip](https://github.com/pytest-dev/pytest-rerunfailures/files/5119984/pytest_rerunfailures-9.1.dev0-py3-none-any.zip) - rename it from `.zip` to `.whl` to use it. I haven't tested it in any way, so use it on your own risk.

No idea when a new release is planned - that's up to the maintainers to decide.

sallner:
@S-Borkowski @The-Compiler I had setup a milestone with some Tickets to be solved before a 9.1 release. In case you definitely need it quickly, I could cut a release tomorrow and move the remaining issues to a 9.2 release.

S-Borkowski:
@sallner If it won't mess with Your work flow much, I would love that!

S-Borkowski:
@sallner Please let me know what was your decision, and when i can expect 9.1 release :)

sallner:
@S-Borkowski @The-Compiler  I have just released version 9.1 of this package. Feel free to use it. https://pypi.org/project/pytest-rerunfailures/9.1/

S-Borkowski:
@sallner <3

PR #115, opened 2020-07-19
refactor the assert_outcomes method to present better messages on assertion failures

gnikonorov:
Currently, an assertion failure in `assert_outcomes` doesn't say much about what's failing. For example, here is what a failure in a non parametrized and parametrized test look like:

```
FAILED test_pytest_rerunfailures.py::test_rerun_report - AssertionError: assert 0 == 99
FAILED test_pytest_rerunfailures.py::test_only_rerun_flag[def test_only_rerun(): raise AssertionError("ERR")-only_rerun_texts0-True] - AssertionError: assert 0 == 99
```

As you can see, the above doesn't tell us what field is incorrect on the `outcomes` object.

With this change, the same failure would look like this:
```
FAILED test_pytest_rerunfailures.py::test_no_rerun_on_pass - AssertionError: outcomes.failed has unexpected value. Expected '99' but got '0'
FAILED test_pytest_rerunfailures.py::test_only_rerun_flag[def test_only_rerun(): raise AssertionError("ERR")-only_rerun_texts0-True] - AssertionError: outcomes.failed has unexpected value. Expected '99' but got '0'
```

The above shows the same error, but with the field that differed included in error message.

### DIFF
--- a/test_pytest_rerunfailures.py
+++ b/test_pytest_rerunfailures.py
@@ -18,6 +18,13 @@ def temporary_failure(count=1):
         count)
 
 
+def check_outcome_field(outcomes, field_name, expected_value):
+    field_value = outcomes.get(field_name, 0)
+    assert field_value == expected_value, \
+            "outcomes.{} has unexpected value. Expected '{}' but got '{}'" \
+            .format(field_name, expected_value, field_value)
+
+
 def assert_outcomes(result, passed=1, skipped=0, failed=0, error=0, xfailed=0,
                     xpassed=0, rerun=0):
     outcomes = result.parseoutcomes()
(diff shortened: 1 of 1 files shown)


test_pytest_rerunfailures.py
```diff
@@ -21,12 +21,17 @@ def temporary_failure(count=1):
...
-    assert outcomes.get('rerun', 0) == rerun
+    check_outcome_field(outcomes, 'passed', passed)
+    check_outcome_field(outcomes, 'skipped', skipped)
+    check_outcome_field(outcomes, 'failed', failed)
+    check_outcome_field(outcomes, 'xfailed', xfailed)
+    check_outcome_field(outcomes, 'xpassed', xpassed)
+    check_outcome_field(outcomes, 'rerun', rerun)
+
+
+def check_outcome_field(outcomes, field_name, expected_value):
```
sallner, line 32: I would like to have the definition before the first usage in this file.

test_pytest_rerunfailures.py
```diff
@@ -21,12 +21,17 @@ def temporary_failure(count=1):
...
+    field_value = outcomes.get(field_name, 0)
+    assert field_value == expected_value, "outcomes.{} has unexpected value. Expected '{}' but got '{}'".format(field_name, expected_value, field_value)
```
sallner, line 34: Please re-format this line to a maximum number of 80 chars.

gnikonorov: Sure, will change as suggested.

On a pseudo aside, do we want to have this enforced in a pre commit rule? Pytest runs [black](https://pypi.org/project/black/) for example, so things like that wouldn’t make it to PR

Let me know, if you’re on board I can raise a PR/issue for it

gnikonorov: Grabbed the item as sugested

gnikonorov:
Thanks for the review @sallner, I've updated as suggested

gnikonorov:
@sallner is this good to merge? I'm not sure what you would consider best practice here since it's 1 approved and 1 reviewed

PR #116, opened 2020-07-19
Check for warnings output during unit tests

gnikonorov:
In the tests, check to make sure that:
- The correct warnings are reported
- The warnings text is correct

Before, we weren't checking but I think we should be since we are setting these warnings in this plugin, and they are not coming from pytest.

No change log as this is trivial

### DIFF
--- a/test_pytest_rerunfailures.py
+++ b/test_pytest_rerunfailures.py
@@ -21,7 +21,7 @@ def temporary_failure(count=1):
 
 
 def assert_outcomes(result, passed=1, skipped=0, failed=0, error=0, xfailed=0,
-                    xpassed=0, rerun=0, warnings=0):
+                    xpassed=0, rerun=0):
     outcomes = result.parseoutcomes()
     assert outcomes.get('passed', 0) == passed
     assert outcomes.get('skipped', 0) == skipped
@@ -29,12 +29,6 @@ def assert_outcomes(result, passed=1, skipped=0, failed=0, error=0, xfailed=0,
     assert outcomes.get('xfailed', 0) == xfailed
     assert outcomes.get('xpassed', 0) == xpassed
     assert outcomes.get('rerun', 0) == rerun
-
-    warnings_key = 'warnings'
-    if warnings <= 1 and pytest_version_gte_53:
-        # due to https://github.com/pytest-dev/pytest/pull/5990
-        warnings_key = 'warning'
-    assert outcomes.get(warnings_key, 0) == warnings
 
 
 def test_error_when_run_with_pdb(testdir):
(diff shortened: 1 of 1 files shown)


icemac:
Hm, the tests fail for `pytest < 5.3`. Do you have any idea why?

The-Compiler:
The problem is that the stored outcomes are `{'failed': 1, 'rerun': 3, 'warnings': 1}` but you're checking for `'warning'` (singular). There were various changes to the plurality there, starting with https://github.com/pytest-dev/pytest/pull/5990 - I'm not sure if `warnings` works on all versions though... Probably worth a try?

gnikonorov:
> The problem is that the stored outcomes are `{'failed': 1, 'rerun': 3, 'warnings': 1}` but you're checking for `'warning'` (singular). There were various changes to the plurality there, starting with [pytest-dev/pytest#5990](https://github.com/pytest-dev/pytest/pull/5990) - I'm not sure if `warnings` works on all versions though... Probably worth a try?

Thanks for the tip @The-Compiler, let me try tonight

gnikonorov:
So I realized what I was doing was inherently wrong. Checking general warnings is extremely fragile at best, since pytest itself emits warnings and there's no way of knowing which warnings came from this plugin vs. pytest. Checking to make sure we emitted at least some amount of warnings 'x' also doesn't tell us anything since those warnings could be coming from somewhere else, so the test would give a false sense of security or just confuse people.

That leads me to think there are 2 possible solutions:
- We keep track of the number of warnings emitted by this plugin in the plugin itself, and compare against that. I personally don't think this is the best idea since we're adding code purely for test purposes, and there's no guarantee that people will maintain this convention
- We do what I updated this PR to do, and just check that stdout contains the warning text we want. I think this is best, since it makes no assumption on the warnings pytest emits, and requires no special maintenance.

Thoughts @icemac and @The-Compiler ?

test_pytest_rerunfailures.py
```diff
@@ -1,4 +1,5 @@
 from unittest import mock
+import pkg_resources
```
icemac, line 2: This import is no longer needed.

test_pytest_rerunfailures.py
```diff
@@ -237,13 +238,17 @@ def test_fail():
     result = testdir.runpytest('--reruns', '3',
                                '--reruns-delay', str(delay_time))
 
+    num_warnings = 0
```
The-Compiler, line 241: This could probably be a `has_warnings` or `warning_expected` boolean now?

gnikonorov: ah yes, this is wrong. sorry for leaving it, let me remove it

The-Compiler: The new version looks better, though now you have two identical `if` statements after each other.

gnikonorov: sorry just changed it

Issue #114, opened 2020-07-16
Add CONTRIBUTING.rst


gnikonorov:
I think we should add a [CONTRIBUTING.rst](https://github.com/pytest-dev/pytest/blob/master/CONTRIBUTING.rst) to this repo to make it clear to newcomers what to add to a successful PR.

I labelled it an enhancement, but it's really documentation.

Issue #80, opened 2019-01-10
Add Flake8 verification


RibeiroAna:
Hi, when I was writing my last contribution I realised that rerunfailures doesn't have Flake8 verifications. I wish I could add that (unless you don't have it for some other reason)

sallner:
Hi @RibeiroAna , thanks for that idea and I would welcome a PR very much. It was mainly, that I did not find the motivation lately. Actually, I would also like to try out `black` for general formatting. But I would prefere that as separate PR maybe in the beginning?

I will just keep a list here with potential maintenance stuff to remind myself and for other to join in. 

- [ ] black as default formatter, travis check for that. Needs some documentation for contributing.
- [ ] Flake8 which complies with the output black generates.
- [ ] Add coverage and activate coveralls.

NathanUrwin:
I would recommend using pylint too. The best case scenario is using pylint and flake8 together, since they both will catch things the other doesn't. Just my two cents!

gnikonorov:
Grabbing this as per https://github.com/pytest-dev/pytest-rerunfailures/pull/115#discussion_r457497214

gnikonorov:
How do we feel about making this repo mirror what pytest is doing and using precommit to enforce this as well

Also, we could use [black](https://pypi.org/project/black/) for code formatting like pytest does

PR #117, opened 2020-07-22
Add in pre-commit to repo

gnikonorov:
This PR adds pre-commit to this plugin and makes it as similar to `pytest`'s  [pre-commit-config](https://github.com/pytest-dev/pytest/blob/master/.pre-commit-config.yaml) as possible. This is intentional to make it easier for developers to switch between projects, and IMO the pytest pre-commit setup is quite nice.

This PR also adds a linting target ( `tox -e linting` ) to this repo for ease of use and applies it to all existing code.

Also, pre commit is run on PR's in the Travis pipeline. This is intentionally redundant, since it is possible for a user to bypass precommit locally and without this CI check we don't have a guaranteed way of enforcing pre-commit. We would still greatly benefit from pre-commit developer side as it gives the user immediate feedback if they are not trying to skip it, and greatly speeds up their development time and cuts down on code review for all involved, since style is already guaranteed to be correct.

The most notable precommit linting addition is [black](https://github.com/psf/black) to enforce code style, and keeps the line length close to @sallner's preferred 80 at 88. See the justification for why [here](https://black.readthedocs.io/en/stable/the_black_code_style.html#line-length).

There are a few outstanding issues that I would like to tackle in follow up PRs, which I will raise once this PR is approved and/or merged. Most notably:
- Support `isolated_build=true` in the tox file ( see [this link](https://tox.readthedocs.io/en/latest/config.html#conf-isolated_build) ). It is deferred since it requires creating a `pyproject.toml` file.
- Add `setup-cfg-fmt` to the pre  commit config like we do in [pytest](https://github.com/pytest-dev/pytest/blob/master/.pre-commit-config.yaml#L39). This requires work on our setup.cfg file, though, and should be deferred since it's beyond scope, and needs consensus IMO before proceeding  
- https://github.com/pytest-dev/pytest-rerunfailures/pull/117#discussion_r462251913

**NOTE**: I did not test the travis config file changes, as I do not know how to. The rest is tested and confirmed working.

Closes https://github.com/pytest-dev/pytest-rerunfailures/issues/114
Closes https://github.com/pytest-dev/pytest-rerunfailures/issues/80

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,7 +5,15 @@ Changelog
 ----------------
 
 - Drop dependency on ``mock``.
+
 - Add in new flag ``--only-rerun`` to allow for users to rerun only certain errors.
+
+- Add support for pre-commit and add a linting tox target.
+  (`#117 <https://github.com/pytest-dev/pytest-rerunfailures/pull/117>`_)
+  (PR from `@gnikonorov`_)
+
+.. _@gnikonorov: https://github.com/gnikonorov
+
 
 9.0 (2020-03-18)
 ----------------
--- a/test_pytest_rerunfailures.py
+++ b/test_pytest_rerunfailures.py
@@ -1,10 +1,11 @@
-from unittest import mock
-import pytest
 import random
 import time
+from unittest import mock
+
+import pytest
 
 
-pytest_plugins = 'pytester'
+pytest_plugins = "pytester"
 
 
 def temporary_failure(count=1):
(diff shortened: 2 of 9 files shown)


gnikonorov:
bump for a review @icemac @sallner

test_pytest_rerunfailures.py
```diff
@@ -15,263 +16,326 @@ def temporary_failure(count=1):
...
+        """
         import pytest
-        @pytest.mark.skipif(reason='{0}')
+        @pytest.mark.skipif(reason='{}')
         def test_skip():
             pass
-    """.format(reason))
-    result = testdir.runpytest('--reruns', '1')
+    """.format(
+            reason
```
sallner, line 65: Those formatings are somehow not so easy to read. Looking forward to f-strings in this case when we stop Supporting Python 2.

And I really do not understand, why this happens. This function call could easily fit on one line. But that is the way you work with black, I guess.

test_pytest_rerunfailures.py
```diff
@@ -385,57 +457,91 @@ def setup_fixture(self):
...
-            ('def test_only_rerun(): raise AssertionError("ERR")', ['ValueError'], False),
-            ('def test_only_rerun(): raise AssertionError("ERR")', [''], True),
-            ('def test_only_rerun(): raise AssertionError("ERR")', ['AssertionError: '], True),
-            ('def test_only_rerun(): raise AssertionError("ERR")', ['AssertionError: ERR'], True),
-            ('def test_only_rerun(): raise AssertionError("ERR")', ['ERR'], True),
-            ('def test_only_rerun(): raise AssertionError("ERR")', ['AssertionError,ValueError'], False),
-            ('def test_only_rerun(): raise AssertionError("ERR")', ['AssertionError ValueError'], False),
-            ('def test_only_rerun(): raise AssertionError("ERR")', ['AssertionError', 'ValueError'], True),
-        ]
+    "file_text, only_rerun_texts, should_rerun",
```
sallner, line 507: The following lines look not so easy to read after black. I wonder, why we actually have the `file_text` in the parameter list here. We should refactor that out afterwards.

tox.ini
```diff
@@ -3,8 +3,17 @@
...
 # and then run "tox" from this directory.
 
+[flake8]
+# NOTE: This is kept in line with Black
+#       See: https://black.readthedocs.io/en/stable/the_black_code_style.html#line-length
+max-line-length = 88
+
 [tox]
-envlist = py{35,36,37,38,py3}-pytest{50,51,52,53,54},
+distshare = {homedir}/.tox/distshare
```
sallner, line 12: According to [docs](https://tox.readthedocs.io/en/latest/config.html#conf-distshare), this seems to be the default, or am I wrong?

tox.ini
```diff
@@ -14,3 +23,12 @@ deps =
     pytest52: pytest==5.2.*
     pytest53: pytest==5.3.*
     pytest54: pytest==5.4.*
+
+[testenv:linting]
+basepython = python3
+commands = pre-commit run --all-files --show-diff-on-failure {posargs:}
+commands_pre =
+    pip install pre-commit
```
sallner, line 31: As I understand tox, this should already be installed, as `pre-commit` is listed in `deps`.

test_pytest_rerunfailures.py
The-Compiler: > Looking forward to f-strings in this case when we stop Supporting Python 2.

What's stopping you? :wink: FWIW you'd also need to drop 3.5 for f-strings, which is currently still [supported upstream](https://devguide.python.org/#status-of-python-branches) for another 2 months (but dropped in various projects already)

sallner: That is true, but was not part of the formatting PR.

test_pytest_rerunfailures.py
gnikonorov: I agree, I'll make a follow up issue to refactor it

test_pytest_rerunfailures.py
gnikonorov: And I really do not understand, why this happens. This function call could easily fit on one line. But that is the way you work with black, I guess.

I tried to reformat the calls, but black still reformatted them. Are you ok with it staying as is?

tox.ini
gnikonorov: it should, you're right. We can remove the entire `commands_pre` section actually

test_pytest_rerunfailures.py
gnikonorov: I added this to the PR description @sallner 
When this PR merges I'll raise the 3 issues in the description

test_pytest_rerunfailures.py
icemac: Let's keep it as it is. This is what happens when using automated tools. The goal when using black was to avoid such formatting discussions in the first place.

sallner: Yeah, this will go away when we use f-strings. We will use black as it is now.

test_pytest_rerunfailures.py
sallner: That sounds good.

CHANGES.rst
```diff
@@ -5,8 +5,16 @@ Changelog
 ----------------
 
 - Drop dependency on ``mock``.
+
 - Add in new flag ``--only-rerun`` to allow for users to rerun only certain errors.
 
+- Add support for pre-commit and add a linting tox target.
+  (`#117 <https://github.com/pytest-dev/pytest-rerunfailures/pull/117>`_)
+  (PR from `@gnikonorov`_)
```
sallner, line 13: :+1:

sallner:
Feel free to merge the PR.

sallner:
I have added two projects for the next feature and major realease. @gnikonorov We should discuss, what tickets you might want to work on for the next feature release. Please add them to the respective milestone. The issues dropping 2.7 and 3.5 support should go into the 10.0 release/milestone.

gnikonorov:
@sallner - when are you considering cutting 9.1?

Issue #108, opened 2020-04-25
pytest_runtest_logreport not called on reruns


christiansandberg:
It seems the `pytest_runtest_logreport` hook is not called for each phase if a previous phase failed. The break statement in the `pytest_runtest_protocol` implementation stops processing the rest of the reports.

icemac:
What would you suggest to do instead?

BeyondEvil:
I just noticed this issue.

Not sure if it is an issue, but it is at least unexpected behaviour.

I maintain `pytest-html` and whilst a "teardown" is executed - it's not reported by the `pytest_runtest_logreport `. This in turn means that the generated report is missing information.

PR #124, opened 2020-08-02
Ensure all test reports are processed regardless of prior report status

gnikonorov:
Ensure that we always process all reports for a test in `pytest_runtest_protocol` regardless of whether a past report for a test failed. We now defer any pre-rerun setup until after all reports are executed while still keeping the same rerun logic.

Fixes https://github.com/pytest-dev/pytest-rerunfailures/issues/108

### DIFF
--- a/pytest_rerunfailures.py
+++ b/pytest_rerunfailures.py
@@ -246,7 +246,7 @@ def pytest_runtest_protocol(item, nextitem):
                 or need_to_rerun
             ):
                 # out of rerun attempts, no errors detected, or we're finishing off
-                # the reports of a filed test log normally
+                # the reports of a failed test, log normally
                 item.ihook.pytest_runtest_logreport(report=report)
             else:
                 # failure detected and reruns not exhausted, since i < reruns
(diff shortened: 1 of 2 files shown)


pytest_rerunfailures.py
```diff
@@ -226,40 +226,45 @@ def pytest_runtest_protocol(item, nextitem):
...
+            out_of_runs = item.execution_count > reruns
             report.rerun = item.execution_count - 1
             xfail = hasattr(report, "wasxfail")
             if (
-                item.execution_count > reruns
+                out_of_runs
                 or not report.failed
                 or xfail
                 or is_terminal_error
+                or need_to_rerun
```
sallner, line 246: With this way of writing we have 5 possible cases in this branch, which will never be properly addressed by any branch coverage. I would prefer a way, so that we need a testcase to make the coverage happy. And I would prefer a testcase.

gnikonorov: This hook was not called for reports of a test run beyond the first failure, which this PR fixes. So, before this PR if a test fails in phase `setup`, its `call` and `teardown` reports aren't processed ( in this case all we do for processing is call `pytest_runtest_logreport`

Now that I look at my change again I have a general question. Why do we not call `pytest_runtest_logreport` for failed test phases when xdist or parellel?

icemac:
What is the current status of this PR?

pyuscherbakov (approved):

BeyondEvil:
Not sure how active @gnikonorov is these days. According to GitHub is last activity was in Jan 202**2**.

I'm happy to pick this up, given the maintainers are willing to have the discussion.

icemac:
@BeyondEvil Thank you for volunteering for this PR. I am here as a maintainer to help and discuss the issue as needed.

Issue #121, opened 2020-07-30
Drop support for python2.7


gnikonorov:
As per https://github.com/pytest-dev/pytest-rerunfailures/pull/117#issuecomment-666177470, we should stop supporting python2.7

Once we drop 2.7 and 3.5, we should also utilize [f-strings](https://www.python.org/dev/peps/pep-0498/)

PR #127, opened 2020-09-26
Reduce pytest requirement to 4.4

pabs3:
The code and tests still work with pytest 4.4

Fixes: commit 43e8c38abe2efb52a03b93a5ea078946fca2a41f

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,7 @@ setup(
     url="https://github.com/pytest-dev/pytest-rerunfailures",
     py_modules=["pytest_rerunfailures"],
     entry_points={"pytest11": ["rerunfailures = pytest_rerunfailures"]},
-    install_requires=["setuptools>=40.0", "pytest >= 5.0"],
+    install_requires=["setuptools>=40.0", "pytest >= 4.4"],
     python_requires=">=3.5",
     license="Mozilla Public License 2.0 (MPL 2.0)",
     keywords="py.test pytest rerun failures flaky",


hugovk:
Should pytest 4.4 also be tested on the CI to ensure it really does work with 4.4?

icemac:
This project currently might run with pytest 4.4 but that would require to test against 4.4 to ensure it.
The current policy is to test against the latest 5 feature versions of pytest. So we decided to drop the older versions to reduce this list at least a bit. That means we still run 25 jobs on TravisCI. 

What is the reasoning behind this PR? Is there a feature you need in pytest 4.4 which is only in a version no longer supporting it?

hugovk:
Also pytest 4.6 was the last version to support Python 2.7: 

* https://docs.pytest.org/en/stable/py27-py34-deprecation.html

Now Python 2.7 is being dropped (https://github.com/pytest-dev/pytest-rerunfailures/issues/121), pytest 5.0 is the oldest version it's possible to support.

pabs3:
I'm packaging pytest-rerunfailures for Debian as part of the dependency
chain for eventually packaging gensim. Unfortunately right now Debian
still only has pytest 4.6 so the pytest dep will have to be changed,
either upstream or in the Debian package. Since version 4.4 still works
I thought probably it didn't need to be bumped in the first place, so I
thought it should be changed upstream. Relaxing it in a Debian patch is
also fine with me if you prefer that.

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

sallner:
@pabs3 Thanks for clarifying your use-case. I really like to restrict the supported versions in this package. If you use pytest 4.6 anyhow and no newer features are used in gensim, please use an older version. There have not been many feature changes since 8.0 so you could use that one, i guess. I will close this PR.

pabs3:
So you suggest using 8.0 in Debian instead of relaxing the dep with 9.1 using a patch?

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

icemac:
Version 8.0 is known to run with pytest 4.4. Newer versions might run but never have been tested against pytest 4.4. So the safest approach would be to use 8.0 against pytest 4.4 instead of patching the current release (9.1).

PR #129, opened 2020-09-26
work around ResultLog support for pytest>=6.1

### LINKED ISSUE #128: _pytest.resultlog was removed in pytest 6.1.0
`pytest-rerunfailures` imports `_pytest.resultlog` but the module was removed in https://github.com/pytest-dev/pytest/commit/ef946d557cc7b7f030805b94c1dbae51f39fcce4 .

As a result, it is no longer possible to use `pytest-rerunfailures` with pytest 6.1.0.

The suggested replacement is [pytest-reportlog](https://github.com/pytest-dev/pytest-reportlog).

ntessore:
This is a workaround that closes #128. It does not address potential integration with the reportlog plugin.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,7 +7,7 @@ Changelog
 Other changes
 +++++++++++++
 
-- Deprecated `--result-log` was removed in pytest 6.1.0.
+- Deprecate ``--result-log``, as it was removed in pytest 6.1.0
 
 
 9.1 (2020-08-26)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,10 @@ Changelog
 9.2 (unreleased)
 ----------------
 
-- Nothing changed yet.
+Other changes
++++++++++++++
+
+- Deprecated `--result-log` was removed in pytest 6.1.0.
 
 
 9.1 (2020-08-26)
--- a/pytest_rerunfailures.py
+++ b/pytest_rerunfailures.py
@@ -4,12 +4,15 @@ import warnings
 
 import pkg_resources
 import pytest
-from _pytest.resultlog import ResultLog
 from _pytest.runner import runtestprotocol
 
 PYTEST_GTE_54 = pkg_resources.parse_version(
     pytest.__version__
 ) >= pkg_resources.parse_version("5.4")
+
+PYTEST_GTE_61 = pkg_resources.parse_version(
+    pytest.__version__
+) >= pkg_resources.parse_version("6.1")
 
 
 def works_with_current_xdist():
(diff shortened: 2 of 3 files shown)


gnikonorov (approved):
As an immediate fix to stop the fatal error, I think this is OK. I will defer to @sallner and/or @icemac for the merge.

@ntessore if this gets merged can you open a separate issue to track integration with `reportlog` so it doesn't get lost?

gnikonorov:
Could you also add a changelog entry to this PR, @ntessore? It can be added in the  [CHANGES.rst](https://github.com/pytest-dev/pytest-rerunfailures/blob/master/CHANGES.rst) file

ntessore:
@gnikonorov I'm really not qualified to open an issue on pytest plugin interoperability, I wouldn't even know if that should be an issue here or in pytest-dev/pytest-reportlog.

CHANGES.rst
```diff
@@ -4,7 +4,10 @@ Changelog
 9.2 (unreleased)
 ----------------
 
-- Nothing changed yet.
+Other changes
++++++++++++++
+
+- Deprecated `--result-log` was removed in pytest 6.1.0.
```
gnikonorov, line 10: Nit: can we change the wording a bit to something like 'Deprecate `--result-log`, as it was removed in  pytest 6.1.0 )'

gnikonorov (changes requested):
Just a small nitpickish comment on the changelog text

gnikonorov:
> @gnikonorov I'm really not qualified to open an issue on pytest plugin interoperability, I wouldn't even know if that should be an issue here or in pytest-dev/pytest-reportlog.

@ntessore I'll open one then. Thank you for fixing the crash

CHANGES.rst
icemac, line 10:
```suggestion
- Deprecate ``--result-log``, as it was removed in pytest 6.1.0
```

nicoddemus (approved):

sallner:
I am unhappy with the fact, that we have now 3 ways of handling the resultlog in this plugin, but it seems that a bugfix release might be appropriate to make various CIs green again.

sallner (approved):
I will merge the changes and cut a release as 9.1.1. We can clean up the rest afterwards.

icemac:
> I will merge the changes and cut a release as 9.1.1. We can clean up the rest afterwards.

Maybe it should become a 9.2 or even 10.0 as the `--result-log` command line option is no linger supported when this plug-in is used together with pytest >= 6.1.

sallner:
@ntessore Thanks for the work, version 9.1.1 is out with the workaround. https://pypi.org/project/pytest-rerunfailures/9.1.1/ Let's see, what we will do with the clean up, and if we need to support resultlog in the future.

hugovk:
Thank you all for the quick fix and release!

PR #135, opened 2020-09-28
Test two newest pytest minor versions

hugovk:
To ensure compatibility with newer versions.

Pytest 6.1.0 currently fails due to https://github.com/pytest-dev/pytest-rerunfailures/issues/128. 

Does this need a CHANGES.rst entry?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,9 @@ env:
   - PYTEST=5.4.1
   - PYTEST=6.0.2
   - PYTEST=6.1.0
+jobs:
+  allow_failures:
+  - env: PYTEST=6.1.0
 install:
   - pip install -q pytest==$PYTEST
   - pip install -q -e .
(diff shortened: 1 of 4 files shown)


icemac:
There should be a change log entry and a maximum version in setup.py until #128 is fixed.

icemac:
Maybe also allowing 6.1 jobs to fail in Travis.

hugovk:
From https://github.com/pytest-dev/pytest-rerunfailures/pull/127#issuecomment-699801977:

> The current policy is to test against the latest 5 feature versions of pytest. So we decided to drop the older versions to reduce this list at least a bit. That means we still run 25 jobs on TravisCI.

With this change, should pytest 5.0 and 5.1 drop off the list?

sallner:
@hugovk Thank you for submitting this PR. As there are so many PRs open partially depending on each other I have added another one #138 which builds on top of the current master. pytest 6.1.0 should be supported this way and we can close this PR.

Issue #122, opened 2020-07-30
Drop support for python3.5


gnikonorov:
As per https://github.com/pytest-dev/pytest-rerunfailures/pull/117#issuecomment-666177470, we should stop supporting python3.5

Once we drop 2.7 and 3.5, we should also utilize [f-strings](https://www.python.org/dev/peps/pep-0498/)

Note that 3.5 reached EOL 2020-09-13 as per [the pyhton developer guide](https://devguide.python.org/#status-of-python-branches). We should not drop support until EOL is reached

PR #137, opened 2020-09-28
Drop support for EOL Python 2.7 and 3.5

hugovk:
Fixes https://github.com/pytest-dev/pytest-rerunfailures/issues/121.
Fixes https://github.com/pytest-dev/pytest-rerunfailures/issues/122.


icemac (changes requested):
I like these changes, they make the code much more readable. Thank you.

The only thing I'd like to request is an entry in the change log to make the PR perfect.

hugovk:
Change log entry added! I also rebased to include the latest change log updates.

sallner (approved):
Thank you for the PR. It seems that putting "code" in f-strings helps with the black formatting very well.

hugovk:
> Thank you for the PR. It seems that putting "code" in f-strings helps with the black formatting very well.

Yeah, stuff like this:

```python
        f"""
        import pytest
        @pytest.mark.flaky(reruns=2)
        def test_pass():
            {temporary_failure(2)}"""
```

In general, I'd avoid putting long or complex things in the `{}` brackets, to help keep things readable. For those, I'd recommend using temporary variables.

sallner (approved):
Thank you, that looks very good!

PR #138, opened 2020-09-28
Test with current pytest versions.

sallner:
This is an exception to rule of the last 5 minor versions of pytest, as this is just the bug fix and a quick workaround. The other versions will be dropped in next minor version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,9 @@ env:
   - PYTEST=5.1.3
   - PYTEST=5.2.4
   - PYTEST=5.3.5
-  - PYTEST=5.4.1
+  - PYTEST=5.4.3
+  - PYTEST=6.0.2
+  - PYTEST=6.1.0
 install:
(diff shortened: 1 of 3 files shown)


hugovk (approved):

gnikonorov (commented):
@sallner was the removal of v5.4.1 from `.travis.yml` intentional?

hugovk:
I think so: 5.4.1 has been replaced with the newest 5.4, that is, 5.4.3.

gnikonorov (approved):
You're right @hugovk, I misread sorry.

LGTM

sallner:
It really takes ages with such a big test matrix.

icemac (changes requested):
I thought the current policy is to support only the latest 5 pytest feature releases. So adding 6.0 and 6.1 our require to drop 5.0 and 5.1. This also reduces the test matrix by 10 test jobs.

icemac (approved):

PR #147, opened 2020-12-12
Add optional condition to marker

BeyondEvil:
The use case we have in [pytest-html](https://github.com/pytest-dev/pytest-html) is that we have a test that is [flaky](https://github.com/pytest-dev/pytest-html/pull/414) only on Windows, so I'd like that test to re-run on the condition that it is running in a Windows environment.

@gnikonorov

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,6 +15,9 @@ Other changes
 - Check for the resultlog by feature and not by version as pytest master does
   not provide a consistent version.
 
+- Add ``condition`` keyword argument to the re-run marker. (Thanks to `@BeyondEvil`_ for the PR)
+
+.. _@BeyondEvil: https://github.com/BeyondEvil
 
 9.1.1 (2020-09-29)
 ------------------
--- a/README.rst
+++ b/README.rst
@@ -99,6 +99,17 @@ You can also specify the re-run delay time in the marker:
       import random
       assert random.choice([True, False])
 
+You can also specify an optional ``condition`` in the re-run marker:
+
+.. code-block:: python
+
+   @pytest.mark.flaky(reruns=5, condition=sys.platform.startswith("win32"))
+   def test_example():
+      import random
+      assert random.choice([True, False])
+
+Note that the test will re-run for any ``condition`` that is truthy.
+
 Output
 ------
 
--- a/pytest_rerunfailures.py
+++ b/pytest_rerunfailures.py
@@ -175,6 +175,16 @@ def get_reruns_delay(item):
     return delay
 
 
+def get_reruns_condition(item):
+    rerun_marker = _get_marker(item)
+
+    condition = True
+    if rerun_marker is not None and "condition" in rerun_marker.kwargs:
+        condition = rerun_marker.kwargs["condition"]
+
+    return condition
+
+
 def _remove_cached_results_from_failed_fixtures(item):
     """
     Note: remove all cached_result attribute from every fixture
@@ -221,6 +231,19 @@ def _should_hard_fail_on_error(session_config, report):
     return True
 
 
+def _should_not_rerun(item, report, reruns):
+    xfail = hasattr(report, "wasxfail")
+    is_terminal_error = _should_hard_fail_on_error(item.session.config, report)
+    condition = get_reruns_condition(item)
+    return (
+        item.execution_count > reruns
+        or not report.failed
+        or xfail
+        or is_terminal_error
+        or not condition
+    )
+
+
 def pytest_runtest_protocol(item, nextitem):
     """
     Note: when teardown fails, two reports are generated for the case, one for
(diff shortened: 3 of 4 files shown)


test_pytest_rerunfailures.py
```diff
@@ -511,3 +511,36 @@ def test_only_rerun_flag(testdir, only_rerun_texts, should_rerun):
     assert_outcomes(
         result, passed=num_passed, failed=num_failed, rerun=num_reruns_actual
     )
+
+
+@pytest.mark.parametrize("condition, expected_reruns", [(1 == 1, 2), (1 == 2, 0)])
```
BeyondEvil, line 516: Maybe `1==1` etc should be just `True` and `False`? Or maybe they should be strings? 🤔

pytest_rerunfailures.py
```diff
@@ -255,6 +269,7 @@ def pytest_runtest_protocol(item, nextitem):
                 or not report.failed
                 or xfail
                 or is_terminal_error
+                or not condition
```
sallner, line 272: This list is getting longer again, this seems like a good place for a refactoring or as an instance of feature creep.

pytest_rerunfailures.py
```diff
@@ -175,6 +175,19 @@ def get_reruns_delay(item):
...
 
+def get_reruns_condition(item):
+    rerun_marker = _get_marker(item)
+
+    condition = True
+    if rerun_marker is not None and "condition" in rerun_marker.kwargs:
+        condition = rerun_marker.kwargs["condition"]
+
+    if not isinstance(condition, bool):
+        warnings.warn("Condition has to be a boolean expression.")
```
sallner, line 186: What is your intend with this warning? Above in the README you propose, that all values evaluating to `True` should be recognized as valid condition.

test_pytest_rerunfailures.py
```diff
@@ -32,7 +32,7 @@ def check_outcome_field(outcomes, field_name, expected_value):
 
 
 def assert_outcomes(
-    result, passed=1, skipped=0, failed=0, error=0, xfailed=0, xpassed=0, rerun=0,
+    result, passed=1, skipped=0, failed=0, error=0, xfailed=0, xpassed=0, rerun=0
```
sallner, line 35: Was this formatting done by black?

test_pytest_rerunfailures.py
sallner: As we have a parametrization anyway here, we could add some more cases. Both mentioned of you seem fine to me. This also serves as kind of documentation after all.

CHANGES.rst
```diff
@@ -15,6 +15,7 @@ Other changes
 - Check for the resultlog by feature and not by version as pytest master does
   not provide a consistent version.
 
+- Add ``condition`` keyword argument to the re-run marker.
```
sallner, line 18: It would be nice, if you could include a link to you github profile.

sallner (changes requested):
Thank you for your contribution. I think, that the addition of this kind of condition is still within the scope of this package, although it overlaps somehow with the possibility to only rerun certain errors. 

The only thing I really want to know is the intention of the warning, otherwise I like the PR.

test_pytest_rerunfailures.py
BeyondEvil: Yes.

```bash
$ black --version
black, version 19.3b0
```

pytest_rerunfailures.py
BeyondEvil: Right, good point! I'm excluding a bunch of values that will evaluate to `True` or `False` here.

Maybe we should leave out the check (and warning) and just rely on the documentation?

pytest_rerunfailures.py
BeyondEvil: One solution is to simply move this out to a `should_rerun`-function. That way there's less cognitive load when understanding either function.

sallner: :+1:

pytest_rerunfailures.py
sallner: I am fine with the documentation .

test_pytest_rerunfailures.py
BeyondEvil: Added more tests.

pytest_rerunfailures.py
BeyondEvil: Removed and changed the documentation.

CHANGES.rst
BeyondEvil: Done. :)

README.rst
```diff
@@ -99,6 +99,17 @@ You can also specify the re-run delay time in the marker:
       import random
       assert random.choice([True, False])
 
+You can also specify an optional ``condition`` in the re-run marker:
+
+.. code-block:: python
+
+   @pytest.mark.flaky(reruns=5, condition=sys.platform.startswith("win32"))
```
BeyondEvil, line 106: should the keyword be `on_condition` instead? 🤔 @sallner

test_pytest_rerunfailures.py
```diff
@@ -511,3 +511,35 @@ def test_only_rerun_flag(testdir, only_rerun_texts, should_rerun):
...
+    "condition, expected_reruns",
+    [
+        (1 == 1, 2),
+        (1 == 2, 0),
+        (True, 2),
+        (False, 0),
+        (1, 2),
+        (0, 0),
+        ("'non-empty'", 2),
+        ("''", 0),
```
BeyondEvil, line 526: it seems both f-strings and `format()` stripped the double quotes, so the single quotes was my solution.

README.rst
sallner: I am fine with `condition` as it is usually used as an enable. Otherwise I would expect something like `omit` or `deactivate`.

test_pytest_rerunfailures.py
sallner: :+1:

sallner (approved):

PR #150, opened 2021-01-30
Adjust parallel check for pytest-xdist `slaveinput` -> `workerinput` renaming

bluetech:
pytest-xdist no longer sets the old name.


sallner (approved):
That looks good and gives us compatibility with `pytest-xdist >= 2.0`. Thank you for your contribution.

PR #151, opened 2021-01-30
Fix SetupState breakage with upcoming pytest 6.3

bluetech:
pytest 6.3 makes some changes to the SetupState class, adjust to that.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -24,6 +24,8 @@ deps =
     pytest54: pytest==5.4.*
     pytest60: pytest==6.0.*
     pytest61: pytest==6.1.*
+    pytest62: pytest==6.2.*
+    pytestmaster: git+https://github.com/pytest-dev/pytest.git@master#egg=pytest
 
 [testenv:linting]
 basepython = python3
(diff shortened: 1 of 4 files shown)


tox.ini
```diff
@@ -24,6 +24,8 @@ deps =
     pytest54: pytest==5.4.*
     pytest60: pytest==6.0.*
     pytest61: pytest==6.1.*
+    pytest62: pytest==6.2.*
+    pytestmaster: git+https://github.com/pytest-dev/pytest.git@master#egg=pytest
```
icemac, line 28: I like testing against `master`.
This change should also be applied to `.github/workflows/test.yml` so CI runs them, too.

bluetech: Sure, added. Using master might of course cause breakage from the pytest side, but hopefully not.

icemac (approved):
Thank you for this PR and improving the test setup for this plug-in.
I like it.

sallner (approved):
I like the changes, too.

icemac:
Thank you for this PR and the review. 😃

PR #144, opened 2020-11-10
Fix higher-scope fixture teardown being executed multiple times on reruns

rogalski:

PR #152, opened 2021-02-15
Bugfix/fixture teardown twice

sallner:
This is a rebase of https://github.com/pytest-dev/pytest-rerunfailures/pull/144

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,7 +33,9 @@ jobs:
             "6.0.*",
             "6.1.*",
             "6.2.*",
-            "master",
+            # We need compatibility with pytest master before we enable it
+            # again.
+            # "master",
         ]
     steps:
       - uses: actions/checkout@v2
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,7 +11,9 @@ Backwards incompatible changes
 
 Bug fixes
 +++++++++
+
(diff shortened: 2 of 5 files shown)


bluetech:
Hi @sallner. I have actually prepared a review of this when I was submitting the other PRs, but didn't complete it. Basically the issue is pretty complicated. Current master is doing entirely the wrong thing (completely clearing the stack), and this PR is not quite right either. I will try to find some time to finish the review.

BTW, in pytest master the `SetupState` is now clearer, stricter and more documented: https://github.com/pytest-dev/pytest/blob/16e21c22a3931bbe33fe2dbc692e149cbbb35b53/src/_pytest/runner.py#L406 might help a bit in the meantime.

sallner:
Hi  @bluetech, I have seen your PRs and contributions to pytest core but did not see a possibility to add a new item after/during the teardown of a fixture to prevent that in case it is still needed, as we do not have the item in `teardown_exact` anymore. Do you think this or a similar approach is still possible in pytest >= 7.0?

icemac:
Is there a new status on this PR? What are the next necessary steps?

BeyondEvil:
Is the issue I'm seeing related to this PR?

Screenshot 2023-07-22 at 03 12 02

Basically for `flaky=N` teardown is logged (possibly also executed) N times extra.

icemac:
As the status of this PR is unclear I am closing it now.

PR #158, opened 2021-04-23
handle worker crash

xoviat:
@sallner What do you think about this?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -61,7 +61,6 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install -U pip
-          python -m pip install pytest-xdist
           if [[ '${{ matrix.pytest-version }}' == 'main' ]]; then
             python -m pip install git+https://github.com/pytest-dev/pytest.git@main#egg=pytest
           else
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -61,7 +61,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install -U pip
-          python -m pip install git+https://github.com/pytest-dev/pytest-xdist.git@7f07d50
+          python -m pip install pytest-xdist
           if [[ '${{ matrix.pytest-version }}' == 'main' ]]; then
             python -m pip install git+https://github.com/pytest-dev/pytest.git@main#egg=pytest
           else
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -61,6 +61,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install -U pip
+          python -m pip install git+https://github.com/pytest-dev/pytest-xdist.git@7f07d50
           if [[ '${{ matrix.pytest-version }}' == 'main' ]]; then
             python -m pip install git+https://github.com/pytest-dev/pytest.git@main#egg=pytest
           else
--- a/pytest_rerunfailures.py
+++ b/pytest_rerunfailures.py
@@ -243,7 +243,7 @@ def _remove_cached_results_from_failed_fixtures(item):
         fixture_defs = fixture_info.name2fixturedefs[fixture_def_str]
         for fixture_def in fixture_defs:
             if getattr(fixture_def, cached_result, None) is not None:
-                result, _, err = getattr(fixture_def, cached_result)
+                result, _, err = getattr(fixture_def, cached_result)  # noqa: F401, F811
                 if err:  # Deleting cached results for only failed fixtures
                     if PYTEST_GTE_54:
                         setattr(fixture_def, cached_result, None)
--- a/pytest_rerunfailures.py
+++ b/pytest_rerunfailures.py
@@ -27,9 +27,16 @@ except ImportError:
     # We have a pytest >= 6.1
     pass
 
+try:
+    from xdist.newhooks import pytest_handlecrashitem
+
+    HAS_PYTEST_HANDLECRASHITEM = True
+    del pytest_handlecrashitem
+except ImportError:
+    HAS_PYTEST_HANDLECRASHITEM = False
+
 
 PYTEST_GTE_54 = parse_version(pytest.__version__) >= parse_version("5.4")
-
 PYTEST_GTE_63 = parse_version(pytest.__version__) >= parse_version("6.3.0.dev")
 
 
@@ -243,7 +250,7 @@ def _remove_cached_results_from_failed_fixtures(item):
         fixture_defs = fixture_info.name2fixturedefs[fixture_def_str]
         for fixture_def in fixture_defs:
             if getattr(fixture_def, cached_result, None) is not None:
-                result, _, err = getattr(fixture_def, cached_result)  # noqa: F401, F811
+                result, _, err = getattr(fixture_def, cached_result)
                 if err:  # Deleting cached results for only failed fixtures
                     if PYTEST_GTE_54:
                         setattr(fixture_def, cached_result, None)
@@ -309,19 +316,16 @@ def pytest_configure(config):
         "between re-runs.",
     )
 
-    try:
-        from xdist.newhooks import pytest_handlecrashitem as _  # noqa: F401, F811
-
+    if HAS_PYTEST_HANDLECRASHITEM:
         if is_master(config):
             config.failures_db = ServerStatusDB()
         else:
             config.failures_db = ClientStatusDB(config.workerinput["sock_port"])
-    except ImportError:
+    else:
         config.failures_db = StatusDB()  # no-op db
 
 
-with suppress(ImportError):
-    from xdist.newhooks import pytest_handlecrashitem as _  # noqa: F401, F811
+if HAS_PYTEST_HANDLECRASHITEM:
 
     def pytest_configure_node(node):
         """xdist hook"""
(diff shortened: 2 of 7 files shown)


xoviat:
Potentially, we can avoid using a database using the `pytest_report_from_serializable` hook in the controller to count failures, and re-configuring the failure history for tests. However, the number of reruns needs to be communicated to the controller before the test run is started. Otherwise, the controller will not be able to determine whether the crash should be re-run or not.

About the db approach: it's as good as it could be, but for the fact that it doesn't work when the nodes are executed on a remote server (as is possible with xdist).

icemac:
This PR fails in all configurations, could you please have a look.

xoviat:
The reason for the failure is that it uses hooks from pytest-xdist.

icemac:
> The reason for the failure is that it uses hooks from pytest-xdist.

So a release of `pytest-xdist` is needed? But on the other hand `pytest-xdist` is not a dependency for `pytest-rerunfailures` and I do not think that it should become one.

xoviat:
I still plan to complete this, but it will be delayed for a bit. The main problem is that the xdist controller needs to know whether a worker crash should be reported as a rerun or a failure, which requires the workers to communicate this information to the controller before running the test setup (so that if the test setup crashes, the controller can rerun it). Currently this is done through a file, but that appears to only work on windows and it doesn't work if the workers are remote. The new plan is set-up a simple key-value database on the controller that's stored in-memory.

xdist will not be a dependency in the final PR, but this functionality will be included if xdist is detected.

icemac:
@xoviat Thank you for the clarification.

xoviat:
new plan: start socket server on controller, use as db:

https://stackoverflow.com/questions/7749341/basic-python-client-socket-example

xoviat:
@icemac I believe this may be close to complete. Do you have any questions about this, or should I clean up the commits?

icemac:
@xoviat The last CI run was a crash. Each and every test failed and had to be killed by GHA after 6 hours. (This does not increase my confidence in this PR.)
What went wrong there and how are you planning to fix it?

xoviat:
The latest run passed (https://github.com/xoviat/pytest-rerunfailures/actions/runs/903077274)

2bff11b wasn't ready yet, I was just pushing changes so that I could see the diffs on github.

.github/workflows/test.yml
```diff
@@ -1,8 +1,8 @@
 name: Test
-
+ 
 on:
   push:
-    branches: [ master ]
+    branches: [ master, crashitem ]
```
icemac, line 5: This change is not needed. To run the PR in this repository (not the forks) I have to approve the run. This was introduced by GitHub to prevent crypto mining.

pytest_rerunfailures.py
```diff
@@ -250,12 +255,155 @@ def _should_not_rerun(item, report, reruns):
...
+def pytest_configure(config):
+    try:
+        from xdist.newhooks import pytest_handlecrashitem as _
+
+        if is_master(config):
+            config.failures_db = ServerStatusDB()
+        else:
+            config.failures_db = ClientStatusDB(config.workerinput['sock_port']) 
+    except ImportError:
+        config.failures_db = StatusDB()  # no-op db
```
icemac, line 271: According to the code of `StatusDB` it is not actually a `no-op` – why?

pytest_rerunfailures.py
```diff
@@ -250,12 +255,155 @@ def _should_not_rerun(item, report, reruns):
...
+    def run_connection(self, conn):
+        with suppress(ConnectionError):
+            while True:
+                op, i, k, v = self._sock_recv(conn).split('|')
+                if op == 'set':
+                    self._set(i, k, int(v))
+                elif op == 'get':
+                    self._sock_send(conn, str(self._get(i, k)))
+
+    def _set(self, i:str, k:str, v:int):
```
icemac, line 378: I do not understand the meaning of the parameters, maybe longer variable names could clarify.

icemac (commented):
I have some questions.

icemac:
Additionally the linter tests fail.

There is one skipped test: Could it be that this is your newly written test?

.github/workflows/test.yml
xoviat: I will revert it. It was used to run it on my own fork (couldn't test it on this one because of the approvals).

pytest_rerunfailures.py
xoviat: If xdist is not detected, then we don't need to record failures across workers, because a test crash `os.exit(1)` will crash pytest. Therefore, this functionality is irrelevant without xdist. With xdist, we need to record failures across workers because we need to reduce the number of times to rerun a test by the number of times that test has crashed a worker.

pytest_rerunfailures.py
xoviat: `i` is a hash of the test name (`test_file.py::test_my_test`), `k` is `f` for failures or `r` for reruns. v is the number of failures or reruns (an int). before we run anything in the test (including setup), the worker needs to record in these in the master so that the master can decide whether the test needs to be rerun if it crashes. the worker must communicate this information because the number of reruns can only be computed by collecting tests, which can only be done in the worker. Again, if xdist is not found this this functionality does nothing.

xoviat:
> There is one skipped test: Could it be that this is your newly written test?

The test is skipped because it requires pytest-xdist development branch. What would you like to do about this?

xoviat:
Again, to clarify, this adds new functionality that reruns a test that crashes a worker with pytest-xdist. If a crash occurs without pytest-xdist, there is nothing that can be done, therefore this code is a no-op without pytest-xdist (the next release as of this writing). However, this functionality is important enough to add because it allows a timed-out test (via pytest-timeout) to be rerun (which cannot be currently done).

pytest_rerunfailures.py
xoviat: To clarify, the StatusDB is an in-memory database stored on the master that records, for each test, the number of failures and the number of reruns. The Status DB is accessed on the worker through a socket connection that binds to a random free port (assigned by the OS). This will allow (eventually) the remote xdist feature to work (when someone requires it).

pytest_rerunfailures.py
icemac: I think I understood the goal but `StatusDB` initializes a socket … maybe there should be an actual no-op StatusDB and a base class for the classes for client and server which encapsulates the common code for these ones. But a socket and probably some other code should not be needed in the no-op version.

icemac:
> The test is skipped because it requires pytest-xdist development branch. What would you like to do about this?

Probably this PR could use that development branch for now until it is merged/released.

pytest_rerunfailures.py
icemac: > `i` is a hash of the test name (`test_file.py::test_my_test`), `k` is `f` for failures or `r` for reruns. v is the number of failures or reruns (an int). before we run anything in the test (including setup), the worker needs to record in these in the master so that the master can decide whether the test needs to be rerun if it crashes. the worker must communicate this information because the number of reruns can only be computed by collecting tests, which can only be done in the worker. Again, if xdist is not found this this functionality does nothing.

It would be important to have this explanation in the code.

pytest_rerunfailures.py
xoviat: Yes, but the socket is not bound to a port. My understanding is that a socket is no-op until it is bound.

xoviat:
@icemac Is this a good summary of required changes?

1. add explanation for set/get.
2. add xdist development branch to tests

icemac:
@xoviat additionally I'd request:

3. fix the linting issues
4. in case of not using xdist make the StatusDB as minimal as possible

xoviat:
Anything else?

icemac:
The linter still fails, see https://github.com/pytest-dev/pytest-rerunfailures/pull/158/checks?check_run_id=2780766837
You can run the linter tests locally by installing `pre-commit`.

xoviat:
@icemac Try again.

xoviat:
cc @icemac

xoviat:
ping @icemac

icemac:
The checks are green now, currently I do not have the time and energy for a final review of this PR. (Maybe @sallner can take over.)
Additionally we need a release of `pytest-xdist` before this PR can be merged.

xoviat:
@icemac Is there an update on this? I'd be willing to resolve the merge conflicts if I could get an indication that this was moving forward.

icemac:
@xoviat Thank you for coming back to this PR. Has the needed xdist release already been created?

xoviat:
Yes

.github/workflows/test.yml
```diff
@@ -61,6 +61,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install -U pip
+          python -m pip install git+https://github.com/pytest-dev/pytest-xdist.git@7f07d50
```
icemac, line 64:
```suggestion
          python -m pip install git+https://github.com/pytest-dev/pytest-xdist.git@7f07d50
```

This line should no longer be necessary or at least install a released version.

icemac:
Additionally there is a problem the linter shows up.

xoviat:
@icemac Fixes appear to work on my branch. To be clear, pytest-xdist is not required with this pr, but it is required if you want to run all of the tests (in this repository, not in general).

icemac (approved):
LGTM.

I will merge it after the CI checks have passed.

icemac (changes requested):
Just before merging  I saw that documentation and a change log entry are missing.

Could you please add them? (in `README.rst` resp. `CHANGES.rst`)

xoviat:
@icemac Done.

icemac (approved):

icemac:
Thank you for this PR. 😃

icemac:
I just created a release including this PR: https://pypi.org/project/pytest-rerunfailures/10.2/

PR #160, opened 2021-05-25
Adapt to pytest's switch to `main` as default branch.

icemac:
Additionally drop support for minor pytest versions which are older than 5 minor releases.


sallner (approved):
Looks good to me, it does not seem, that there are any other code changes necessary.

PR #162, opened 2021-06-18
condition can be evaluated in the real time

15klli:
use the skipping.evaluate_condition from pytest in order to keep the same behavior with pytest.mark.skipif. It evaluates the condition in real time so that we can decide whether the case should be rerun in runtime

Change-Id: I2575d368c79480223c84498513a4ef605db0c576

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -108,6 +108,20 @@ You can also specify an optional ``condition`` in the re-run marker:
       import random
       assert random.choice([True, False])
 
+You can use ``@pytest.mark.flaky(condition)`` similarly as ``@pytest.mark.skipif(condition)``, see `pytest-mark-skipif <https://docs.pytest.org/en/6.2.x/reference.html#pytest-mark-skipif>`_
+
+.. code-block:: python
+
+    @pytest.mark.flaky(reruns=2,condition="sys.platform.startswith('win32')")
+    def test_example():
+        import random
+        assert random.choice([True, False])
+    # totally same as the above
+    @pytest.mark.flaky(reruns=2,condition=sys.platform.startswith("win32"))
+    def test_example():
+      import random
+      assert random.choice([True, False])
+
 Note that the test will re-run for any ``condition`` that is truthy.
 
 Output
--- a/pytest_rerunfailures.py
+++ b/pytest_rerunfailures.py
@@ -1,9 +1,14 @@
+import os
+import platform
 import re
+import sys
 import time
+import traceback
 import warnings
 
 import pkg_resources
 import pytest
+from _pytest.outcomes import fail
 from _pytest.runner import runtestprotocol
 
 HAS_RESULTLOG = False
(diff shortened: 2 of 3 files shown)


icemac:
@15klli Thank you for this PR. It currently fails when running the tests, could you please fix it.

15klli:
> @15klli Thank you for this PR. It currently fails when running the tests, could you please fix it.

ok, I will fix them.

15klli:
I fix them by coping the `evaluate_condition ` from `skipping.py(python3.8.10)`, and all check passed in my PC. And I have to remind that this feature has compatible problem like the following:
```
@pytest.mark.flaky(condition="non_empty",reruns=2)
def test_demo():
     assert False
```
The above testcase will rerun 2 times(condition means  True) with the front version but will failed and get error with the version having this feature, since it doesn't define `non_empty`.

But I believe almost no one will write "non_empty" string instead of writing  `True` directly so that the case will always rerun if failed.

15klli:
@icemac Hello, can you review it when you are free?

icemac:
@15klli Currently the linter complains. Could you please fix these issues, too?

15klli:
> @15klli Currently the linter complains. Could you please fix these issues, too?

Done.

icemac (approved):
LGTM.

icemac:
Thank you for this PR. 😃

sallner:
Sorry for being late to the party. But there is still a changelog entry missing. And with this change in the condition parsing, it actually is a backwards incompatible change. @15klli what is the prospect of the copied code? As far as I understand the refactoring in https://github.com/pytest-dev/pytest/pull/7388 it is include in pytest versions >= 6.0. That would be something to note at the code.

15klli:
> Sorry for being late to the party. But there is still a changelog entry missing. And with this change in the condition parsing, it actually is a backwards incompatible change. @15klli what is the prospect of the copied code? As far as I understand the refactoring in [pytest-dev/pytest#7388](https://github.com/pytest-dev/pytest/pull/7388) it is include in pytest versions >= 6.0. That would be something to note at the code.
 
The reasons why I copied the code instead of referring to the `skipping.evaluate_condition`  are that:
1. The low version pytest doesn't have such function
2. The higher version pytest will throw a exception if the marker doesn't have the `reason` value which is required by `skipif`  marker

That's why copied the code though I know it is not a graceful way.
Does `That would be something to note at the code`  mean that  I should write  the note into the python code?

icemac:
I added a change log entry in https://github.com/pytest-dev/pytest-rerunfailures/commit/138db6cc15c0f66ef12156d2680a39c8c5c80af2.

15klli:
@icemac when will you package it and upload to `pypi.org` ? I want to apply this feature on my project.

icemac:
@15klli I just released https://pypi.org/project/pytest-rerunfailures/10.1/

icemac:
@sallner Could you please fix the still open issues directly on `master`? Sorry for being too quick to merge the PR.

PR #168, opened 2021-09-15
Add support for Python 3.10

hugovk:
Python 3.10.0 final is due for release in October:

* https://www.python.org/dev/peps/pep-0619/

The final release candidate is now out and the Python release team has issued a call to action for community members:

> We strongly encourage maintainers of third-party Python projects to prepare their projects for 3.10 compatibilities during this phase. As always, report any issues to the Python bug tracker.

https://discuss.python.org/t/python-3-10-0rc2-is-now-available/10496?u=hugovk

So let's also test and add support for 3.10.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -38,7 +38,7 @@ jobs:
           "main",
         ]
         exclude:
-          # Only pylast 6.2+ supports Python 3.10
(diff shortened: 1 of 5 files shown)


hugovk:
Thanks for the typo fix!

icemac (approved):
LGTM.

Thank you for this PR. 😃

PR #194, opened 2022-10-18
Expose `reruns` and `reruns_delay` through `pytest.ini` file

laky55555:


icemac:
Thank you for this PR. It currently breaks many of the currently supported combinations of Python versions and pytest versions. Could you please take a look into the test failures and suggest a solution (There is #196, so I am open to drop support for old versions if needed.)

laky55555:
Hey, thanks for the hint, I added a small change to be compatible with older versions of pytest

icemac:
@laky55555 Could you please rebase your PR onto the current `main` branch and move the change log entry to entry for the new release.

laky55555:
Hi @icemac, not sure if I understand the procedure.
When you say rebase PR onto the current `main` do you mean rebase and squash or just rebase as it is already ahead of `main`.
Regarding the change log, I moved the changes under `10.4` and added the release date for `10.3`

icemac:
@laky55555 `rebase` either means `git rebase` or just to merge the current upstream `main` branch into your branch and fixing the change log so the entry belongs to the not yet released version.

icemac (approved):

icemac:
Thank you for reviewing this PR. 😃

PR #201, opened 2023-01-10
fix: don't bind all ips

sileht:
Binding all IPs is kind of Security issues and also very annoying when
the OS has a firewall blocking application doing such thing or opening a
popup to allow it (like the default MacOS or Windows firewall).

The client uses "localhost", so the server can only bind the ip
associated with the hostname "localhost".

This will avoid opening this tools to anyone on the network and avoid
having to configure windows or macos firewall to make pytest run.

### DIFF
--- a/pytest_rerunfailures.py
+++ b/pytest_rerunfailures.py
@@ -440,7 +440,7 @@ class SocketDB(StatusDB):
 class ServerStatusDB(SocketDB):
     def __init__(self):
         super().__init__()
-        self.sock.bind(("", 0))
+        self.sock.bind(("localhost", 0))
         self.sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
 
         self.rerunfailures_db = {}


icemac (approved):
LGTM.

icemac:
In #202 I am fixing the main branch, so this PR should run successfully afterwards.

icemac:
Thank you for this PR. 😃

icemac:
Published this fix in https://pypi.org/project/pytest-rerunfailures/11.0/.

sileht:
Thx! That was fast 😀

codefromthecrypt:
would you be open to changing this from localhost -> 127.0.0.1 to rule out problems due to misconfiguration or resolution? That would be simpler than exposing a configuration to choose how it is bound, but wanted to ask y'all which is a better way out?

codefromthecrypt:
and to be clear, I would for posterity change both client and server bind IPs to whichever (hard-coded 127.0.0.1 or configurable single BIND_HOST or pytest arg)

icemac:
@codefromthecrypt I'd be open for a pull request to for the suggested change.

PR #204, opened 2023-01-12
Add tests for PyPy 3.9 and update GitHub actions

oscargus:
Not sure how to specify these for `tox` though...

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,7 +27,7 @@ jobs:
           "3.9",
           "3.10",
           "3.11",
-          "pypy-3.7",
+          "pypy-3.9",
         ]
         pytest-version: [
           "6.0.*",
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -28,6 +28,8 @@ jobs:
           "3.10",
           "3.11",
           "pypy-3.7",
+          "pypy-3.8",
+          "pypy-3.9",
```
icemac, line 32: I see no need to test on multiple PyPy versions. Testing on one should be enough.
Why do you think it is necessary?

icemac (changes requested):
Besides the PyPy tests I like the changes.

.github/workflows/test.yml
oscargus: One may consider testing on one of 3.8 and 3.9 as the most recent versions of PyPy do no longer support 3.7. https://www.pypy.org/posts/2022/12/pypy-v7311-release.html

The 3.7 test is running PyPy 7.3.9, while 3.8 and 3.9 is running 7.3.11. (Which explains the tox thing, so should be possible to get one in there as well, it was not clear how to distinguish them, but that is because it is not required...)

(I was chasing down an issue in the Matplotlib test suite when executed with PyPy, but I am now quite sure that it didn't come from here.)

So to conclude, I'd suggest testing of one of pypy-3.8 and pypy-3.9 to test with the latest version (it was not so clear to me that one should be enough when I did this). Plus then pypy-3.7 if you want to keep that.

icemac: I'd suggest to replace pypy-3.7 with pypy-3.9 then.

oscargus: Updated actions for the lint workflow as well.

icemac (approved):

icemac:
Thank you for this PR. 😃

PR #205, opened 2023-02-07
Run teardown once after reruns

### LINKED ISSUE #51: Fixtures marked (scope='session') are run on every re-run
I would expect that session fixtures are run once per session - but it seems if a `flaky` test reruns, it will run session scoped fixtures' teardown code, and then their setup again.

This can result in costly setup/teardown being unnecessarily rerun, and also seems to go against the whole point of labelling the fixture `session`

Reproduction:

```Python
@pytest.fixture(scope="session")
def example_fixture():
    print("\nSetup")
    yield
    print("Teardown")

@pytest.mark.flaky(reruns=1)
def test_example_fixture(example_fixture):
    assert 0
```

Running with `pytest -s -v ...` I get:

```
plugins: timeout-1.2.0, rerunfailures-2.2, repeat-0.4.1, ordering-0.5
collected 1 items 

scripts/autotest/tests/conftest.py::test_example_fixture 
Setup
Teardown
RERUN
scripts/autotest/tests/conftest.py::test_example_fixture 
Setup
Teardown
FAILED
```

lukasNebr:
Fixes #51 

Teardown of session, class, ... scoped fixture will be executed once after rerunning tests.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,11 @@ Changelog
 11.1 (unreleased)
 -----------------
 
+Bug fixes
++++++++++
+
+- Run teardown of session, class, ... scoped fixtures only once after rerunning tests
+
 Features
 ++++++++
 
--- a/pytest_rerunfailures.py
+++ b/pytest_rerunfailures.py
@@ -12,6 +12,7 @@ from contextlib import suppress
 
 import pytest
 from _pytest.outcomes import fail
+from _pytest.python import Function
 from _pytest.runner import runtestprotocol
 from packaging.version import parse as parse_version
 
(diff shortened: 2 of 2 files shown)


icemac:
Thank you for this PR. Could you please additionally write a test which shows that the bug is now fixed?

icemac:
Additionally tests seem to fail in `pytest < 7`. We currently support `pytest >= 6.1`.

lukasNebr:
I added a test to verify that the bug is fixed. 

The implementation should now be compatible with `pytest >= 6.1`.  I have forgotten the versions lower than 7 before. Sorry for that.

icemac:
Thank you for fixing the code and adding a test.

sallner (approved):
Looks good from my perspective.

icemac:
Thank you for (reviewing) this PR. 😃

icemac:
I just released https://pypi.org/project/pytest-rerunfailures/11.1/ containing this fix.

icemac:
These changes seem to cause #206. @lukasNebr Do you have an idea what could cause that error?

lukasNebr:
I have no idea whats the cause of this error. I tried to reproduce the behavior but I was not successful. 

Basically the `pytest_runtest_protocol` hook starts the setup, calls the test method and executes the teardown. As a part of teardown the `pytest_runtest_teardown` hook will be executed.

As you can see in the code the `execution_count` for an item is set (line 563) before the setup is executed (line 572). So in the teardown the attribute should exist. 

https://github.com/pytest-dev/pytest-rerunfailures/blob/b20fef5333ce1bf9b327df41cfa9a3f7b282ee5c/pytest_rerunfailures.py#L562-L573

If you have also no idea what the cause of the error is I would implement a workaround to avoid the error.

icemac:
Regarding to #206 the problem is caused if another plugin also defines `pytest_runtest_protocol`. Let's move the discussion how to solve this problem over there.

Issue #206, opened 2023-02-10
AttributeError: 'TestCaseFunction' object has no attribute 'execution_count'


wangzelin007:
Hi pytest admin,
We run all tests using pytest.
But recently we upgraded pytest-rerunfailures to latest (version 11.1) , which caused all tests failed.
This is the error log:
![image](https://user-images.githubusercontent.com/18628534/218044196-50eafb88-7fb7-4ef9-b056-34baa73cd6a5.png)
After I locked pytest-rerunfailures to 11.0 in this [PR](https://github.com/Azure/azure-cli/pull/25419), all tests came back to work.

icemac:
Thank you for your bug report, it seems to be caused by #205 which added the code line where the error happens.

rokm:
This is likely caused by having installed another plugin that overrides `pytest_runtest_protocol` as well; so `pytest` ends up running the other plugin's function, and `execution_count` ends up missing on the item.

 In our case, it seems to be the combination of `pytest-rerunfailures` and `flaky` that is triggering the problem (and although we were playing roulette with which plugin's `pytest_runtest_protocol` gets run, there were apparently compatible enough that the problem went unnoticed).

wangzelin007:
Hi @rokm,
Here's all the softwares we install, but we don't use [flaky](https://pypi.org/project/flaky/):
- MarkupSafe-2.1.2
- Pygments-2.14.0
- alabaster-0.7.13
- argcomplete-2.0.0
- astroid-2.8.6
- attrs-22.2.0
- azdev-0.1.40
- azure-common-1.1.28
- azure-core-1.26.3
- azure-multiapi-storage-1.0.0
- babel-2.11.0
- cachetools-5.3.0
- certifi-2022.12.7
- cffi-1.15.1
- chardet-5.1.0
- charset-normalizer-3.0.1
- colorama-0.4.6
- cryptography-39.0.1
- distlib-0.3.6
- docutils-0.19
- exceptiongroup-1.1.0
- execnet-1.9.0
- filelock-3.9.0
- flake8-4.0.1
- gitdb-4.0.10
- gitpython-3.1.30
- idna-3.4
- imagesize-1.4.1
- iniconfig-2.0.0
- isodate-0.6.1
- isort-5.12.0
- jinja2-3.1.2
- jmespath-1.0.1
- knack-0.10.1
- lazy-object-proxy-1.9.0
- mccabe-0.6.1
- msrest-0.7.1
[...]

PR #207, opened 2023-02-16
Fix crash in runtest teardown hook

lukasNebr:
This PR refers to issue #206 

I improved the implementation of the pytest runtest hook. Now the hook is compatible with test cases that use the TestCase class as base class. This fixes [this error](https://github.com/pytest-dev/pytest-rerunfailures/issues/206#issuecomment-1430305454).

In addition I added a check to avoid [this attribute error](https://github.com/pytest-dev/pytest-rerunfailures/issues/206#issue-1579261105). It is clear that this does not fix the general problem with hooks beeing overwritten by other plugins.


CHANGES.rst
```diff
@@ -1,7 +1,16 @@
...
 
-11.1 (unreleased)
+11.2 (unreleased)
+-----------------
+
+Bug fixes
++++++++++
+
+- Fix crash during teardown when runtest protocol hook is overwritten by another plugin
+- Fix crash during teardown when TestCase class is used as base class
```
icemac, line 11:
```suggestion
- Fix crash during teardown when runtest protocol hook is overwritten by another plugin.
- Fix crash during teardown when TestCase class is used as base class.
```

CHANGES.rst
```diff
@@ -1,7 +1,16 @@
 Changelog
 =========
 
-11.1 (unreleased)
+11.2 (unreleased)
+-----------------
```
icemac, line 5:
```suggestion
11.1.1 (unreleased)
-------------------
```

pytest_rerunfailures.py
```diff
@@ -513,23 +512,29 @@ def pytest_runtest_teardown(item, nextitem):
         # flaky
         return
 
+    if hasattr(item, "execution_count") is False:
```
icemac, line 515:
```suggestion
    if not hasattr(item, "execution_count"):
```

This should be enough I see no need to check for identity with `False`.

icemac (approved):
LGTM, just some little changes.

Thank you for this PR.

icemac:
Released in https://pypi.org/project/pytest-rerunfailures/11.1.1/

PR #226, opened 2023-06-30
Fix handling of --rerun-except arguments

### LINKED ISSUE #225: Multiple --rerun-except arguments are ANDed instead of ORed
While working on #221, I noticed that when passing multiple --rerun-except arguments (or, now when passing a list to `@pytest.mark.flaky(rerun_except=[...])`), all failures are rerun unless they match _all_ of the patterns.

With this test:
``` python
def test_fail():
    raise AssertionError()
```

Running with the example --rerun-except args from the readme gives:
```
$ pytest --reruns 5 --rerun-except AssertionError --rerun-except OSError
======================== test session starts =========================
platform linux -- Python 3.9.2, pytest-7.4.0, pluggy-1.2.0
rootdir: /tmp/test_rerun
plugins: rerunfailures-11.1.2
collected 1 item

test_rerun.py R                                                [100%]R [100%]R [100%]R [100%]R [100%]F [100%]

============================== FAILURES ==============================
_____________________________ test_fail ______________________________

    def test_fail():
>       raise AssertionError()
E       AssertionError

test_rerun.py:2: AssertionError
====================== short test summary info =======================
FAILED test_rerun.py::test_fail - AssertionError
===================== 1 failed, 5 rerun in 0.02s =====================
```

I think the intention in this case (as implied by the docs) is that the test should not be rerun, since the failure matches one of the --rerun-except arguments.

Currently the logic in [_should_hard_fail_on_error](https://github.com/pytest-dev/pytest-rerunfailures/blob/bfc030f841ba5c81493c87ba2a3db3339b500df1/pytest_rerunfailures.py#L318C7-L318C7) is:
- If neither --only-rerun nor --rerun-except are used, all failures are rerun
- Else, if one or more --only-rerun args is given, a failure is rerun if it matches any --only-rerun pattern
- Else, if one or more --rerun-except args is given, a failure rerun if it **fails to match any** --rerun-except pattern

The fix for this isn't exactly obvious, as the intended relationship between --only-rerun and --rerun-except is a bit unclear when they are used together.  My proposal would be:
- If neither --only-rerun nor --rerun-except are used, all failures are rerun (no change)
- Else, if --only-rerun is used alone, a failure is rerun if it matches any --only-rerun pattern (no change)
- Else, if --rerun-except is used alone, a failure is rerun if it matches _none_ of the --rerun-except patterns (new)
- Else, if --only-rerun and --rerun-except are used together, a failure is only rerun if it matches at least one --only-rerun pattern, AND _none_ of the --rerun-except patterns (new)

cfulljames:
Previously, a failure had to match all --rerun-except arguments to prevent a test from rerunning.  Now it will not rerun if one or more are matched.

Additionally, if --only-rerun and --rerun-except are used together, a failure must match at least one --only-rerun pattern and none of the --rerun-except patterns for the test to be rerun.  Previously a test would be rerun if it matched neither --only-rerun nor --rerun-except.

Fixes #225

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,8 +19,13 @@ Features
 Bug fixes
 +++++++++
 
--- Fix handling of ``--rerun-except`` and ``--only-rerun`` when used together,
-   and multiple instances of ``--rerun-except``.
+-- Failures are now rerun only if they match at least one ``--only-rerun``
+   pattern (if given) and none of the ``--rerun-except`` patterns. Previously,
+   using both ``--only-rerun`` and ``--rerun-except`` together could cause
+   failures to be rerun even if they did not match any ``--only-rerun``
+   pattern, and when using multiple ``--rerun-except`` patterns, all failures
+   would be rerun unless they matched every pattern.
+   (`#225 <https://github.com/pytest-dev/pytest-rerunfailures/issues/225>`_)
 
 
 11.1.2 (2023-03-09)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,11 @@ Features
 
 - Add ``only_rerun`` and ``rerun_except`` arguments to ``@pytest.mark.flaky`` marker.
 
+Bug fixes
++++++++++
+
+-- Fix handling of ``--rerun-except`` and ``--only-rerun`` when used together,
+   and multiple instances of ``--rerun-except``.
 
 11.1.2 (2023-03-09)
 -------------------
--- a/pytest_rerunfailures.py
+++ b/pytest_rerunfailures.py
@@ -315,6 +315,24 @@ def _get_rerun_filter_regex(item, regex_name):
     return regex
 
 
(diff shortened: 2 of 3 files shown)


CHANGES.rst
```diff
@@ -9,6 +9,11 @@ Features
 
 - Add ``only_rerun`` and ``rerun_except`` arguments to ``@pytest.mark.flaky`` marker.
 
+Bug fixes
++++++++++
+
+-- Fix handling of ``--rerun-except`` and ``--only-rerun`` when used together,
+   and multiple instances of ``--rerun-except``.
```
icemac, line 16: This bugfix changes the behaviour, so could you please write about the behaviour before and after the fix here as you did in the PR description?

icemac (changes requested):
I like the fix, I'd just request a bit more documentation in the change log as this is what users of the package might read.

icemac (approved):
Thank you for fixing this issue!

PR #240, opened 2023-10-20
Fix Issue #239

metheuspsc:
Fixes issue #239 and adds test coverage

PR #244, opened 2023-11-21
Fix crash with setup errors and --rerun-except flag

Joonalai:
This PR fixes internal errors on setup (non-existent fixtures etc) that occur when using --rerun-except flag.

I am wondering whether reruns should occur when failure happens on setup part of the test?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,7 @@ Changelog
 - Fix crashitem names mismatch between client and server.
   (`#172 <https://github.com/pytest-dev/pytest-rerunfailures/issues/172>`_)
 
+- Fix crash when setup fails with --rerun-except flag.
 
 12.0 (2023-07-05)
 -----------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,7 @@ Changelog
   (`#172 <https://github.com/pytest-dev/pytest-rerunfailures/issues/172>`_)
 
 - Fix crash when setup fails with --rerun-except flag.
+  (`#230 <https://github.com/pytest-dev/pytest-rerunfailures/issues/230>`_)
 
 12.0 (2023-07-05)
 -----------------
--- a/pytest_rerunfailures.py
+++ b/pytest_rerunfailures.py
@@ -278,8 +278,12 @@ def _matches_any_rerun_error(rerun_errors, report):
 
 def _matches_any_rerun_except_error(rerun_except_errors, report):
     for rerun_regex in rerun_except_errors:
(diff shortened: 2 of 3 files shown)


test_pytest_rerunfailures.py
```diff
@@ -617,6 +617,26 @@ def test_rerun_except_and_only_rerun(
...
+    testdir.makepyfile(
+        """
+        import pytest
+
+        @pytest.fixture()
+        def fixture_setup_fails(non_existent_fixture):
+            return 1
+
+        def test_will_not_run(fixture_setup_fails):
+            assert fixture_setup_fails==1"""
```
icemac, line 630:
```suggestion
            assert fixture_setup_fails == 1"""
```

icemac (approved):
LGTM besides some styling suggestion.

icemac:
> I am wondering whether reruns should occur when failure happens on setup part of the test?

If an error occurs that early probably rerunning does not help, but this could be solved in another PR.

Joonalai:
> > I am wondering whether reruns should occur when failure happens on setup part of the test?
> 
> If an error occurs that early probably rerunning does not help, but this could be solved in another PR.

I agree, maybe another issue or PR should be opened for this.

pytest_rerunfailures.py
```diff
@@ -282,6 +282,17 @@ def _try_match_reprcrash(rerun_errors, report):
...
+    for rerun_regex in rerun_except_errors:
+        try:
+            if re.search(rerun_regex, report.longrepr.reprcrash.message):
+                return True
+        except AttributeError:
+            if re.search(rerun_regex, report.longreprtext):
+                return True
+    return False
+
+
```
icemac, line 295:
```suggestion
```

This issue was already fixed in #240. But I'd like to keep the test case.

CHANGES.rst
```diff
@@ -7,6 +7,7 @@ Changelog
 - Fix crashitem names mismatch between client and server.
   (`#172 <https://github.com/pytest-dev/pytest-rerunfailures/issues/172>`_)
 
+- Fix crash when setup fails with --rerun-except flag.
```
icemac, line 10:
```suggestion
- Fix crash when setup fails with --rerun-except flag.
  (`#230 <https://github.com/pytest-dev/pytest-rerunfailures/issues/230>`_)
```

icemac (changes requested):

icemac (approved):

icemac:
Thank you for this PR. 😃

pytest_rerunfailures.py
Joonalai: Hah, bad timing it seems :D

PR #250, opened 2024-02-21
precommit: update min `py37`

Borda:
the minimal package supported Python version is 3.7 so it makes sense to also align all used liters

https://github.com/pytest-dev/pytest-rerunfailures/blob/5f8fab9b551cf96b3f30930e63d9f19eb8a076d3/setup.cfg#L43

cc: @sallner @florianpilz @hugovk

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,7 +4,7 @@ repos:
       rev: 23.11.0
       hooks:
           - id: black
-            args: [--safe, --quiet, --target-version, py36]
+            args: [--safe, --quiet, --target-version, py37]
     - repo: https://github.com/pre-commit/pre-commit-hooks
       rev: v4.5.0
       hooks:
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -4,7 +4,7 @@ repos:
       rev: 23.11.0
       hooks:
           - id: black
-            args: [--safe, --quiet, --target-version, py36]
+            args: [--safe, --quiet, --target-version, py37]
```
hugovk, line 7: Please could you also switch to use `https://github.com/psf/black-pre-commit-mirror` for the repo? It's a bit faster.

And we might as well update them too: `pre-commit autoupdate --jobs 0`

Borda: Well Black di some significant changes so I may suggest doing it in a separate PR #252

hugovk: I don't see any changes from Black in https://github.com/pytest-dev/pytest-rerunfailures/pull/252 :)

hugovk (approved):

PR #254, opened 2024-02-21
Drop support for EOL Python 3.7 and bump GitHub Actions

hugovk:
EOL since 2023-06-27.

image

https://devguide.python.org/versions/

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -2,9 +2,7 @@ name: Lint
 
 on:
   push:
-    branches: [ master ]
   pull_request:
-    branches: [ master ]
     type: [ "opened", "reopened", "synchronize" ]
 
 env:
@@ -16,10 +14,10 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       - name: Cache
-        uses: actions/cache@v3
+        uses: actions/cache@v4
         with:
           path: |
             ~/.cache/pip
@@ -30,7 +28,7 @@ jobs:
             lint-v1-
 
       - name: Set up Python
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
           python-version: '3.x'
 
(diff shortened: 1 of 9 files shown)


.github/workflows/lint.yml
```diff
@@ -16,10 +14,10 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
```
Borda, line 17: Can we add deependabot to keep track of these changes?

Borda (approved):
lest add to the PR name also bump all GH actions

hugovk:
Thanks for the review!

To merge, also "At least 1 approving review is required by reviewers with write access".

setup.cfg
```diff
@@ -1,6 +1,3 @@
-[bdist_wheel]
-universal = 0
-
```
icemac, line 3: Are you sure about this? It was added to prevent creating wheels which claim Python 2 support in their name.

icemac (commented):
I like the changes but are not sure about one change. When it is cleared it can give the necessary approval.

setup.cfg
hugovk: 0 is the default.

To confirm, here's the result of `python -m build` on `master`:

```
pytest-rerunfailures-13.1.dev0.tar.gz           pytest_rerunfailures-13.1.dev0-py3-none-any.whl
```

and with this PR:

```
pytest-rerunfailures-13.1.dev0.tar.gz           pytest_rerunfailures-13.1.dev0-py3-none-any.whl
```

Neither has `py2` in the filename.

icemac: That's nice to hear, so we can remove this setting from all other repos, too. Thank you.

icemac (approved):
I like the changes!

setup.cfg
hugovk: Tiny PRs, or roll it into other updates (for example, removing 3.7, or adding 3.12) :)

Borda: Happy to help if you tell me which projects...

PR #262, opened 2024-02-25
Fix tests reported as failures when rerun happened due to exception raised from fixture teardown when using only_rerun

### LINKED ISSUE #261: only_rerun with exception raised from fixture teardown reruns test but report previous runs as failure
Hi.
I have a need to do some checks in a fixture teardown to determine whether a test should rerun or not.
Because rerunfailures only allows filtering based on exception, what I had in mind is that to check whatever I want in the fixture and raise the custom exception so the test will rerun.
However, while this package detects a rerun is needed when processing the teardown report, the call (test) report was already processed and it was not marked to be rerun, because the rerun_only condition didn't match the test exception, as the custom exception was raised from the fixture.

Eventually, the test does re-run until it pass, but pytest still report the test failure outcomes.

Note that the reruns works just fine when not using only_rerun condition.

Example code:
```
import random

import pytest

class MyCustomError(Exception):
    pass

@pytest.fixture
def rerun_custom_checker(request):
    yield

    if request.node.rep_call.outcome == "passed":
        return

    raise MyCustomError()


@pytest.mark.flaky(reruns=10, reruns_delay=1, only_rerun=["MyCustomError"])
def test_example(rerun_custom_checker):

    i = random.randint(2, 4)
    if i != 3:
        assert False
```

image

What I had in mind, is to pre-process the reports and set a flag whether rerun is needed, if it was found to be needed for setup/call/teardown.
Then, when processing the reports as usual, and just check for this flag.

Something like:
image

I can raise a PR with that change, assuming this solution is acceptable (It works fine when testing this change)

shlompy:
Fixes https://github.com/pytest-dev/pytest-rerunfailures/issues/261


icemac:
@shlompy Thank you for your PR. Currently many tests break. Could you please have a look what's causing this?

hugovk:
Thanks for the review request, but I'll stick to general CI/maintenance things for this library and leave the internals to others 👍

shlompy:
resolved conflicts

shlompy:
gentle reminder, can this fix please be reviewed?

tests/test_pytest_rerunfailures.py
```diff
@@ -472,7 +472,7 @@ def pytest_runtest_teardown(item):
             assert item.execution_count == 3"""
     )
     result = testdir.runpytest("--reruns", "2")
-    assert_outcomes(result, passed=3, rerun=2)
+    assert_outcomes(result, passed=1, rerun=2)
```
icemac, line 475: The test name says that `execution_count` is exposed to the test runner. So it should be `3` passed tests, as this is set during tear down.

I do not know why this behaviour is here and why it breaks now.

src/pytest_rerunfailures.py
```diff
@@ -548,9 +548,24 @@ def pytest_runtest_protocol(item, nextitem):
...
+        for r in reports:
+            rerun_except_errors = _get_rerun_filter_regex(item, "rerun_except")
+            except_found = rerun_except_errors and _matches_any_rerun_except_error(
+                rerun_except_errors, r
+            )
+            if except_found:
+                break
+
+        should_not_rerun = True
+        for r in reports:
```
icemac, line 563: The for loop in line 567 uses `for report in reports`, so I'd like it here to for consistency.

icemac (changes requested):
It is not clear why the test breaks, it looks like a smell.

I'd like to see a test like you had in the issue failing without the changes and succeeding with them.

src/pytest_rerunfailures.py
shlompy: thanks, fixed and also fix except reports loop to break after finding first except match so later report without except match will not override it (This fix all the test failures)

icemac:
@shlompy Are you still interested in this PR or can I close it as abandoned?

Issue #267, opened 2024-03-28
pytest >= 8.2 is not yet supported (not only tests fail)


icemac:
See https://github.com/pytest-dev/pytest-rerunfailures/actions/runs/8409152908/job/23026172136

This ~~can be a temporary problem or a~~ is a hint that we have to change something here to make it compatible again.

mgorny:
This is now affecting pytest 8.2.0.

Chris-Peterson444:
This looks like a repeat of what https://github.com/pytest-dev/pytest-rerunfailures/commit/0ab54f00d68bb4feadae58d2f4e30f6663495571 tried to address (the teardown fixtures are not being called correctly). I spent some time debugging this this afternoon and the logic still looks correct, although it is effectively the same error. I'm not too familiar with the pytest internals but I wonder if the fix is somewhere along the lines of keeping a reference to the original setupstate instead of the depending on the final tests setupstate object?

dkuchynskyi:
Hi guys, I think we've faced the same problem using pytest-rerunfailures with pytest 8.2.2 and pytest-xdist 3.6.1. we run tests with xdist loadscope and rerunfailures and on the step of collecting tests it fails:

```
platform linux
pytest-8.2.2
configfile: pytest.ini
plugins: rerunfailures-14.0, xdist-3.6.1
created: 3/3 workers
3 workers [43 items]

scheduling tests via LoadScopeScheduling
INTERNALERROR> def worker_internal_error(
INTERNALERROR>         self, node: WorkerController, formatted_error: str
INTERNALERROR>     ) -> None:
INTERNALERROR>         """
INTERNALERROR>         pytest_internalerror() was called on the worker.
INTERNALERROR>     
INTERNALERROR>         pytest_internalerror() arguments are an excinfo and an excrepr, which can't
INTERNALERROR>         be serialized, so we go with a poor man's solution of raising an exception
INTERNALERROR>         here ourselves using the formatted message.
INTERNALERROR>         """
INTERNALERROR>         self._active_nodes.remove(node)
INTERNALERROR>         try:
INTERNALERROR> >           assert False, formatted_error
```

We had to rollback to pytest 7.4.4 but left rerunfailures-14.0, xdist-3.6.1 and now it works properly.
Could you look at it please
thanks

mkmoisen:
[...]

PR #276, opened 2024-09-21
Adds `--fail-on-flaky` option.

filiplajszczak:
Fails the test run with custom exit code when test passes on rerun.

Some tests are failing on `pytest-main` and it looks like it's due to #267 and unrelated to the PR.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,8 +4,11 @@ Changelog
 14.1 (unreleased)
 -----------------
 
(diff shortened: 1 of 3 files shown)


icemac:
Thank you for this PR. I think the feature is valuable.

I'd like to wait with a merge until #267 is fixed to be sure it does not increase the work to be done there.

icemac (approved):

PR #278, opened 2024-10-31
fix compatibility with pytest 8.2 by restoring deleted finalizers

### LINKED ISSUE #267: pytest >= 8.2 is not yet supported (not only tests fail)

jakkdl:
Fixes #267

The problem in #267 arose from https://github.com/pytest-dev/pytest/pull/11833, which made pytest not add finalizers for cached fixtures.
`pytest_runtest_teardown` relied on cached fixtures getting their finalizers repeatedly re-added, as it deleted all(?) finalizers on every re-run. This PR makes sure to save away those finalizers, so we can add them back when we're done rerunning and let them execute.

I suspect the current solution will break with xdist or something, given that `suspended_finalizers` is a global dict. So it'd probably be more prudent to do something with the database?

This is my first time looking at pytest-rerunfailures, so I'd love if any current maintainers wanted to have opinions or even finish up the PR now that I've mostly figured out the messy part.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,8 @@ Changelog
 14.1 (unreleased)
 -----------------
 
-- Nothing changed yet.
+- Fix compatibility with pytest 8.2.
+  (`#267 <https://github.com/pytest-dev/pytest-rerunfailures/issues/267>`_)
 
 
 14.0 (2024-03-13)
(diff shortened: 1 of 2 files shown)


icemac (approved):
I like the change. As it fixes our tests it is definitly a step forward.

icemac:
Thanbk you very much for digging into this issue.

PR #279, opened 2024-11-05
Update versions

### LINKED ISSUE #267: pytest >= 8.2 is not yet supported (not only tests fail)

icemac:
- Drop support for Python 3.8
- Drop support for pytest < 7.4
- Add support for Python 3.13
- Add support for pytest 8.2 and 8.3.

Fixes #267

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,7 @@ ignore = [".pre-commit-config.yaml"]
 
 [project]
 name = "pytest-rerunfailures"
-version = "14.1.dev0"
+version = "15.0.dev0"
 description = "pytest plugin to re-run tests to eliminate flaky failures"
 dynamic = [
   "readme",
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -46,7 +47,7 @@ classifiers = [
 ]
 dependencies = [
   "packaging>=17.1",
-  "pytest>=7.2",
+  "pytest>=7.4",
```
asottile-sentry, line 50:
```suggestion
  "pytest>=7.4,!=8.2.2",
```

as per https://github.com/pytest-dev/pytest-rerunfailures/issues/267#issuecomment-2457891221

pyproject.toml
```diff
@@ -38,6 +38,7 @@ classifiers = [
   "Programming Language :: Python :: 3.10",
   "Programming Language :: Python :: 3.11",
   "Programming Language :: Python :: 3.12",
+  "Programming Language :: Python :: 3.13",
```
hugovk, line 41: Also remove the 3.8 classifier above and bump `requires-python`.

hugovk (approved):

pyproject.toml
icemac: Good catch!

icemac:
Thank you for reviewing this PR. 😃

PR #281, opened 2024-11-20
Release 15.0

icemac:


sallner (approved):
LGTM. We have two times 'pytest' as keyword in the pyproject.toml, but that is a different issue.

icemac:
> LGTM. We have two times 'pytest' as keyword in the pyproject.toml, but that is a different issue.

I created #282 to fix this.

PR #283, opened 2024-11-20
Initial draft docs

filiplajszczak:
As there were no docs for the project apart from the README, I have created 
basic draft sphinx documentation. It is partially based on including sections 
from the README, but I have also added some additional information in more 
structured way.

The questions are where to host it and if the README should be kept in the 
current extensive form or if it should be shortened and the rest just kept 
in the docs. What is the right balance between the README and the docs?

I temporarily [hosted the draft docs on  PythonAnywhere](https://pytest-rerunfailures-draft-docs-fjl.pythonanywhere.com/) so it could be quickly viewed rendered.

The choice of theme and other bikeshedding can be done later, but I think 
it's a good start.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@
 bin/
 build/
 dist/
+docs/_build/
 include/
 lib/
 pip-selfcheck.json
--- a/README.rst
+++ b/README.rst
@@ -1,8 +1,12 @@
 pytest-rerunfailures
 ====================
 
+.. START-SHORT-DESCRIPTION
+
 pytest-rerunfailures is a plugin for `pytest <https://pytest.org>`_ that
 re-runs tests to eliminate intermittent failures.
+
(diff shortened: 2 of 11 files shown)


icemac:
@filiplajszczak Hi! I really like this documentation. Thank you very much!
I'd suggest to host it on readthedocs.org and add a badge to the readme. (I can do both once the PR is merged.)

The PR is still in draft mode, what do you think is still needed to make it ready?

filiplajszczak:
Yeah, there is some potential extra work, but there always is. Nothing that could not happen in the next iteration. So maybe let's start with what we have without gold plating.

Wouldnl you be able to setup something with RTD and make it part of CI and release processes?

icemac:
Let me try whether I have enough powers in the pytest organization to do so.

icemac (approved):

PR #284, opened 2024-11-22
Configure for readthedocs

icemac:
The result looks like this: https://pytest-rerunfailures--284.org.readthedocs.build/284/

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,15 @@
+# Read the Docs configuration file for Sphinx projects
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Set the OS, Python version and other tools you might need
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.12"
+
+# Build documentation in the "docs/" directory with Sphinx
+sphinx:
+  configuration: docs/conf.py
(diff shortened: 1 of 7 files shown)


.readthedocs.yaml
```diff
@@ -0,0 +1,15 @@
...
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Set the OS, Python version and other tools you might need
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.12"
```
hugovk, line 11: Can we use newer?
```suggestion
  os: ubuntu-24.04
  tools:
    python: "3.13"
```
Or latest?
```suggestion
  os: ubuntu-lts-latest
  tools:
    python: "3"
```

webknjaz:
I recommend wiring Sphinx via tox to avoid maintaining two integrations.

Here's how you call it then: https://github.com/ansible/awx-plugins/blob/e613d5a/.readthedocs.yaml.

icemac:
@webknjaz This looks way more complicated than what we currently have.

webknjaz:
> @webknjaz This looks way more complicated than what we currently have.

It's necessary if we don't want to have the workflow of building locally for contributors to be "go read the source code and attempt to guess which command to run, where, in what env and how".

icemac:
@webknjaz I got your point and I added tox environment to build the docs locally. But I am not going to use it at readthedocs. Using their workflow seems way easier.

icemac:
I do not plan to improve this PR any further.
So I enabled auto-merge and started waiting for a review.
If someone wants to adopt this PR for further improvements, I can disable auto-merge again.

hugovk (approved):
Thanks!

PR #288, opened 2025-02-13
Adjust `--fail-on-flaky` behavior to improve exit status handling

### LINKED ISSUE #287: --fail-on-flaky not compatible with mark usage
`--fail-on-flaky` only exits with code 7 if `--reruns` is passed via CLI. Otherwise it breaks the execution regardless of flakiness.

Also the implementation implies if I add `--reruns` i want this to always return exit status 7... this was not what I was expecting.

test_flaky.py:
```python
@pytest.mark.flaky(reruns=10)
def test_flaky_test():
    assert True
```

```sh
$ pytest -k test_flaky_test --fail-on-flaky
...
File "$HOME/.local/share/virtualenvs/<venv>/lib/python3.9/site-packages/pytest_rerunfailures.py", line 632, in pytest_sessionfinish
    if session.config.getvalue("reruns") > 0:
TypeError: '>' not supported between instances of 'NoneType' and 'int'
```

ricardo-passthrough:
Enhance the `--fail-on-flaky` option to set a custom exit status when tests are retried and fail. This change ensures that the exit status reflects the presence of flaky tests more accurately. Additional tests validate the new behavior.

Fixes #287

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,12 @@ Changelog
 15.1 (unreleased)
 -----------------
 
-- Nothing changed yet.
+Bug fixes
++++++++++
+
+- Fix ``--fail-on-flaky`` option to fail the test run with custom exit code
+  only when reruns are detected.
+  (`#287 <https://github.com/pytest-dev/pytest-rerunfailures/issues/287>`_)
 
 
 15.0 (2024-11-20)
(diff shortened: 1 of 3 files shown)


icemac:
Thank you for this pull request. Unfortunately it fails the tests. Could you please have a look?

icemac (approved):

PR #289, opened 2025-03-23
Use SPDX license identifier

cdce8p:
[PEP 639](https://peps.python.org/pep-0639/) added support for SPDX license identifier.
https://spdx.org/licenses/MPL-2.0.html

Metadata diff
```diff
 ...
-License: Mozilla Public License 2.0 (MPL 2.0)
+License-Expression: MPL-2.0
 ...
-Classifier: License :: OSI Approved :: Mozilla Public License 2.0 (MPL 2.0)
 ...
 License-File: LICENSE
 ...
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,6 +28,6 @@ repos:
         - id: pyproject-fmt
           additional_dependencies: [ tox ]
     - repo: https://github.com/abravalheri/validate-pyproject
-      rev: v0.16
+      rev: v0.24.1
       hooks:
         - id: validate-pyproject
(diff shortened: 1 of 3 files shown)


icemac:
I like the change but I think it is a bit too early to require setuptools >= 77. I think it is to mature a bit more and needs more uptake.

cdce8p:
> I like the change but I think it is a bit too early to require setuptools >= 77. I think it is to mature a bit more and needs more uptake.

Are there any specifics you'd like to see from setuptools before adopting it or just that it's too new of a version. Unfortunately, `v77` is the first release that fully supports license expressions with the "new" syntax.

An alternative would be to re-use `license.text` with the SPDX identifier for the time being.

icemac (approved):
Looks reasonable.

Issue #292, opened 2025-04-23
only_rerun="AssertionError" doesn't trigger reruns when AssertionError is raised by built-in assert statement


yugokato:
I wanted my test to rerun only when it fails with an `AssertionError` to limit the scope of reruns. However, I found out that `only_rerun="AssertionError"` doesn't work when the `AssertionError` is raised by built-in `assert` statement.
It does work if I explicitly raise the error using `raise AssertionError()` but that's not how assertions are typically written in `pytest`.

- assert (doesn't work)
```
$ pytest 
========================= test session starts ==========================
<snip>                                     

test_something.py F                                              [100%]

=============================== FAILURES ===============================
____________________________ test_something ____________________________

    @pytest.mark.flaky(only_rerun="AssertionError")
    def test_something():
>       assert False
E       assert False

test_something.py:7: AssertionError
======================= short test summary info ========================
FAILED test_something.py::test_something - assert False
========================== 1 failed in 0.02s ===========================
```

- AssertionError (works)
```
$ pytest 
========================= test session starts ==========================
<snip>

test_something.py RF                                             [100%]

=============================== FAILURES ===============================
____________________________ test_something ____________________________

    @pytest.mark.flaky(only_rerun="AssertionError")
    def test_something():
>       raise AssertionError()
E       AssertionError

test_something.py:6: AssertionError
======================= short test summary info ========================
FAILED test_something.py::test_something - AssertionError
====================== 1 failed, 1 rerun in 0.03s ======================
```

Environment:
pytest==8.3.5
pytest-rerunfailures==15.0

yugokato:
I've just looked at the source code out of curiosity. I think we could change to pass the `call` object to [`_should_hard_fail_on_error()`](https://github.com/pytest-dev/pytest-rerunfailures/blob/15.0/src/pytest_rerunfailures.py#L280-L305) at [this line](https://github.com/pytest-dev/pytest-rerunfailures/blob/15.0/src/pytest_rerunfailures.py#L534), then access the actual error directly via `call.excinfo.type.__name__` to compare it with the value specified as `only_rerun`, instead of using `report.longrepr.reprcrash.message` or `report.longreprtext` in [here](https://github.com/pytest-dev/pytest-rerunfailures/blob/15.0/src/pytest_rerunfailures.py#L269C5-L277). 
Matching against the actual exception would be more reliable than using `report.longrepr.reprcrash.message` or `report.longreprtext`, which can cause a false positive.

PR #293, opened 2025-04-23
Update exception filtering to match against actual exception

yugokato:
Fixes https://github.com/pytest-dev/pytest-rerunfailures/issues/292

This PR aims to change [this function](https://github.com/pytest-dev/pytest-rerunfailures/blob/15.0/src/pytest_rerunfailures.py#L269-L277) to check specified `only_rerun` and `rerun_except` value(s) against the actual exception, instead of using `report.longrepr.reprcrash.message` or `report.longreprtext`.

I wasn't sure why `report.longrepr.reprcrash.message` and `report.longreprtext` are currently used. So please disregard/close the PR if my change looks invalid or would break something.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,12 @@ Bug fixes
 
 - Switched to using the SPDX license identifier in the project metadata.
 
+Other changes
++++++++++++++
+
+- Change exception filtering logic to match ``AssertionError`` raised via
+  ``assert`` statements when filtering by "AssertionError".
+  (`#292 <https://github.com/pytest-dev/pytest-rerunfailures/issues/292>`_)
 
 15.0 (2024-11-20)
 -----------------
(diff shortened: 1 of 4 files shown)


icemac (approved):
LGTM, could you please add a change log entry describing your change.

yugokato:
> LGTM, could you please add a change log entry describing your change.

Thanks, I'll do. Which changelog category should I use for this? I initially thought the original behavior was a bug, but given the test coverage and README note around this, it seems it was an intentional design choice. 
Also I'm going to remove [this README section](https://github.com/pytest-dev/pytest-rerunfailures/blob/15.0/README.rst?plain=1#L110-L115) as it is no longer valid with my change.

yugokato:
@icemac I've added my change log under the "Other Changes" category for now. Please let me know if you prefer a different one.

icemac:
LGTM, enabling automerge.

icemac:
Released in https://pypi.org/project/pytest-rerunfailures/15.1/.

Issue #274, opened 2024-09-11
[BUG] --reruns breaks yielded fixture


GeorgeAtKistler:
I have a session scoped fixture which should be run before and after the test session:
```python
import logging
import pytest

@pytest.fixture(scope='session', autouse=True)
def my_session_fixture():
    logging.error("################ Pre ################")
    yield    
    logging.error("################ Post ################")
       

def test_dummy():
    assert 1 == 0
```

This usually shows the following log:
```python
pytest --log-cli-level=INFO tests.py
============================================== test session starts ==============================================
platform linux -- Python 3.10.12, pytest-8.3.3, pluggy-1.5.0
rootdir: /home/me
plugins: rerunfailures-14.0
collected 1 item

tests.py::test_dummy
------------------------------------------------ live log setup -------------------------------------------------
ERROR    root:tests.py:6 ################ Pre ################
FAILED                                                                                                    [100%]
----------------------------------------------- live log teardown -----------------------------------------------
ERROR    root:tests.py:8 ################ Post ################


=================================================== FAILURES ====================================================
_________________________________________________ test_dummy __________________________________________________

    def test_dummy():
>       assert 1 == 0
E       assert 1 == 0

tests.py:12: AssertionError
---------------------------------------------- Captured log setup -----------------------------------------------
ERROR    root:tests.py:6 ################ Pre ################
--------------------------------------------- Captured log teardown ---------------------------------------------
ERROR    root:tests.py:8 ################ Post ################
============================================ short test summary info ============================================
FAILED tests.py::test_dummy - assert 1 == 0
=============================================== 1 failed in 0.16s ===============================================
```

How ever, if I use the `--reruns` parameter, the 2nd part (after the yield) does not get executed (`################ Post ################` never gets called):
```python
pytest --log-cli-level=INFO --reruns 1 tests.py
============================================== test session starts ==============================================
platform linux -- Python 3.10.12, pytest-8.3.3, pluggy-1.5.0
rootdir: /home/me
plugins: rerunfailures-14.0
collected 1 item

tests.py::test_dummy
------------------------------------------------ live log setup -------------------------------------------------
ERROR    root:tests.py:6 ################ Pre ################
RERUN                                                                                                     [100%]
tests.py::test_dummy FAILED                                                                               [100%]

=================================================== FAILURES ====================================================
__________________________________________________ test_dummy ___________________________________________________

    def test_dummy():
>       assert 1 == 0
E       assert 1 == 0

[...]

PR #294, opened 2025-04-24
Upper bound for pytest version, since >= 8.2 is not supported

chaporgin:
According to the following PRs [^1], [^2] version 8.2 is not supported and that's what we encountered while trying to use fixtures with yield statement on pytest 8.3.5. Downgrading seemingly fixed the problem.

[^1]: https://github.com/pytest-dev/pytest-rerunfailures/issues/272#issuecomment-2348152479
[^2]: https://github.com/pytest-dev/pytest-rerunfailures/issues/274#issuecomment-2348149775

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,7 +44,7 @@ classifiers = [
 ]
 dependencies = [
   "packaging>=17.1",
-  "pytest!=8.2.2,>=7.4",
+  "pytest<8.2,>=7.4",
 ]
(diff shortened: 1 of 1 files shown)


The-Compiler:
#267 is fixed by #278, so this seems incorrect to me.

webknjaz:
Also, it's not a good idea to have upper bounds in libs. Besides, the depresolver would backtrack to the previous version a decide to install something potentially even more problematic.

chaporgin:
Made a comment into the linked issue, since it persists.

PR #305, opened 2025-09-02
Change 'localhost' to '127.0.0.1'

Cyperwu:
In some weird case like a misconfigured VPN application, localhost would be resolved as "localhost.somedomain", which leading to an `OSError: [Errno 49] Can't assign requested address`.

I think we should replace "localhost" with a more explicit "127.0.0.1".


Cyperwu:
Sure.

PR #307, opened 2025-09-18
Add a `--force-reruns` to override rerun count globally

### LINKED ISSUE #306: Please make it possible to override the rerun count globally
Consider the following sample:

```python
import pytest


@pytest.mark.flaky
def test_foo():
    assert False


@pytest.mark.flaky(reruns=3)
def test_bar():
    assert False
```

This causes the tests to be rerun exactly once and 3 times, independently of `--reruns=...` parameter on the CLI. I would find it really helpful to be able to override the upstream defaults somehow via the command-line, as our downstream workflows often suffer from much more flakiness than the pristine upstream CI.

mgorny:
Add a new `--force-reruns` command-line option that can be used to override the rerun count globally, irrespectively of individual test markers.

Fixes #306.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,9 @@ Changelog
 
 - Changed "localhost" to "127.0.0.1" to avoid bad hostname resolution
 
+- Added ``--force-reruns`` to override rerun count globally.
+  Fixes `#306 <https://github.com/pytest-dev/pytest-rerunfailures/issues/306>`_.
+
 16.0.1 (2025-09-02)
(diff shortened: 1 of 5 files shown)


icemac:
@mgorny Thank you for the PR let me prepare a PR that fixed the test issue.

icemac:
@mgorny The GHA failure is now fixed upstream. Could you please have a look into the Copilot comments?

mgorny:
I'm not a native speaker, so I guess the statistical data set knows better.

PR #311, opened 2025-10-02
New `--all-reruns-need-to-pass` argument

msaelices:
This addresses your requirement to verify that non-deterministic tests (that work ~90% of the time) pass consistently by requiring all reruns to pass after an initial failure.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,8 +6,7 @@ Changelog
 
 - Added the ``pytest_rerunfailures_rerun_policy`` hook: plugins can return a
   ``RerunPolicy`` to give a specific failure class its own rerun count /
-  all-reruns-need-to-pass mode and an outcome tag (``report.rerun_tag``), without
-  changing how other failures are rerun.
+  all-reruns-need-to-pass mode, without changing how other failures are rerun.
 
 
 16.3 (2026-05-22)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,10 @@ Changelog
 16.4 (unreleased)
 -----------------
 
-- Nothing changed yet.
+- Added the ``pytest_rerunfailures_rerun_policy`` hook: plugins can return a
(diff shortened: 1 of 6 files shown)


icemac:
Thank you for your PR. Unfortunately I do not understand which problem it solves could you please explain your use case with an example?

Additionally the merge conflict and lint errors could be solved.

icemac:
@msaelices Are you still interested in this PR, otherwise I'll close it as abandoned.

msaelices:
> @msaelices Are you still interested in this PR, otherwise I'll close it as abandoned.

We are using it to run evals that are not deterministic, like LLM based

Scenario: Let's say you have 100 evals with, e.g. 5% of failure each one of them, which is ok for you. 

Even with a seemingly "good" 5% per-eval failure rate, running 100 evals makes it very likely, with a ~99.4% chance, that at least one fails.

How do you differentiate between a harmless flaky failure and an actual regression, where the failure rate jumps from something acceptable like 5% to, for example, 50%?

We use `--reruns` together with `--all-rerun-need-to-pass` to handle this.

For example, with `--reruns 20` and `--all-rerun-need-to-pass`, we can ensure that the observed failure rate for that test remains around ~5%, instead of allowing a much higher regression to slip through unnoticed.

icemac:
@msaelices Did I understand correctly? When using `--all-rerun-need-to-pass` a failing test is not only rerun until it succeeds but until the defined number of reruns. It is allowed to only failf the first time and has to succeed all the follwing re-runs.

msaelices:
> @msaelices Did I understand correctly? When using `--all-rerun-need-to-pass` a failing test is not only rerun until it succeeds but until the defined number of reruns. It is allowed to only failf the first time and has to succeed all the follwing re-runs.

Exactly. It enforces that all follow-up reruns must pass for the test to be considered successful.

icemac:
@msaelices Sorry for letting this PR lay around for so long. I now has merge conflicts which cannot be fixed via the web UI, could you please fix them locally.

PR #312, opened 2025-10-10
Update to current versions of pre-commit hooks.

icemac:


sallner (approved):
Looks good to me, it does not seem, that a changelog entry would be really necessary.

PR #329, opened 2026-05-19
Add --rerun-show-tracebacks to surface retried failures

### LINKED ISSUE #156: Please add option to show failure messages for all reruns
**Problem:** 
pytest-rerunfailures plugin shows only assertion message from the last launched test case. From the code snippet, it is tough to identify what went wrong on the first and second attempts.
   
    ~  pytest test_one.py --reruns 2 -v   
 
    test_one.py::test_fail RERUN    [100%]
    test_one.py::test_fail RERUN    [100%]
    test_one.py::test_fail FAILED   [100%]

    ================================ FAILURES ==================================
    ________________________________ test_fail _________________________________

    def test_fail():
    >       assert False, 'Fake message'
    E       AssertionError: Fake message
    E       assert False

    test_one.py:5: AssertionError

    ============================== short test summary info ======================
    FAILED test_one.py::test_fail - AssertionError: Fake message
    ================================ 1 failed, 2 rerun in 0.13s =================
    
It would be extremely useful (especially in debugging) to see all previous assertion messages such as described in the FAILURES section for the last failed test case. For instance, the last assertion message does not mean that all previous tests failed with the same reason.

**The solution I would like to see:** 
I would like to see some extra argument that will allow launching logs for all failed/rerun testcases. As far as I see from the code below failure logging is omitted at all (log=False): https://github.com/pytest-dev/pytest-rerunfailures/blob/3f8685e4bb92659fb3eab76aa624ad1244ca9770/pytest_rerunfailures.py#L276 
It would be great if a new argument will allow to manage `log` parameter so that user will take control on logging failures.

I know that we can extend test summary with `-r rR` option, but unfortunately, it is not enough.

vokracko:
Currently `pytest-rerunfailures` only emits the traceback for the final attempt of a flaky test. If the last attempt passes, the failures that triggered the reruns are silently discarded; if the last attempt fails, the earlier tracebacks are gone even when they hit a different exception. That makes triaging flaky tests harder than it needs to be — you can see that a test was retried, but not what actually went wrong on each attempt.

This adds `--rerun-show-tracebacks`, which appends each retried attempt's traceback to the `rerun test summary info` section. The section is emitted automatically when the flag is set, so `-rR` is no longer required.

For example, a test that hits a transient assertion twice and passes on the third attempt:

```
test_demo.py::test_eventually_passes RERUN                               [100%]
test_demo.py::test_eventually_passes RERUN                               [100%]
test_demo.py::test_eventually_passes PASSED                              [100%]
=========================== rerun test summary info ============================
RERUN test_demo.py::test_eventually_passes
>       assert attempt > 2, f"transient failure on attempt {attempt}"
E       AssertionError: transient failure on attempt 1
E       assert 1 > 2
RERUN test_demo.py::test_eventually_passes
>       assert attempt > 2, f"transient failure on attempt {attempt}"
E       AssertionError: transient failure on attempt 2
E       assert 2 > 2
```

As a side effect, the per-char loop in `pytest_terminal_summary` is gone. The loop called `show_rerun` once per matching char in `reportchars`, so passing `-r rR` doubled every RERUN line in the summary; the new code calls `show_rerun` once. Happy to revert that part if the duplication is intentional.

Fixes #156.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,14 @@ Changelog
 16.3 (unreleased)
 -----------------
 
-- Nothing changed yet.
+Features
+++++++++
+
+- Add ``--rerun-show-tracebacks`` option to display tracebacks from failed
+  attempts that were retried, including tests that eventually passed. The
+  rerun summary section is emitted automatically when the flag is set, so
+  ``-rR`` is no longer required to see the tracebacks.
+  Fixes `#156 <https://github.com/pytest-dev/pytest-rerunfailures/issues/156>`_.
 
 
 16.2 (2026-05-13)
--- a/README.rst
+++ b/README.rst
@@ -208,6 +208,22 @@ of re-runs specified in test markers, pass ``--force-reruns``:
 
    $ pytest --force-reruns 5
 
+Show tracebacks for retried failures
+------------------------------------
+
+By default only the *final* attempt of a flaky test produces a traceback,
+so failures from earlier attempts (including those of tests that ultimately
+pass after a rerun) are silently discarded. To inspect them, pass
+``--rerun-show-tracebacks``:
+
+.. code-block:: bash
+
+   $ pytest --reruns 2 --rerun-show-tracebacks
+
+Each retried attempt's traceback is appended to the ``rerun test summary
+info`` section. The section is emitted automatically when the flag is set,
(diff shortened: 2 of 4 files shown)


icemac (approved):
LGTM.

PR #338, opened 2026-07-15
fix: prevent rerun when teardown error matches --rerun-except

### LINKED ISSUE #270: rerun-except option captures only first error of the test, ignoring teardown error
Hello!

I have a problem, related to using rerun-except.
Scenario that causes the problem:

- test run with --rerun-except SomeError
- test uses function scoped fixture
- error in test occures, for example: assert fail
- teardown of function scoped fixture is triggered
- SomeError is raised in the teardown
- test reruns, even though SomeError was raised

Expected behaviour: test should fail without rerun

It seems to me that only the first raised error in test is taken in account when using --rerun-except?
Is there any way to fix it through configuration or is it a library issue?

pctablet505:
Fixes #270

When --rerun-except is used and a teardown failure matches the excluded error, the test was still rerun because only the current report's stage was checked for terminal errors. Now all stages of the current execution are considered, so a terminal teardown error prevents reruns as expected.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,7 +12,7 @@ Bug fixes
   was considered, so a ``--rerun-except``-matching error raised during
   teardown was ignored and the test was rerun anyway.
   Fixes `#270 <https://github.com/pytest-dev/pytest-rerunfailures/issues/270>`_.
-  
+
 Features
 ++++++++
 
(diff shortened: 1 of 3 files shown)


icemac (approved):
LGTM. Thank you for the PR.

PR #343, opened 2026-08-14
Fix regression making --pdb unusable without configured reruns

### LINKED ISSUE #342: `ERROR: --reruns incompatible with --pdb` in version 16.5 even if not setting `--reruns`
### `ERROR: --reruns incompatible with --pdb` in version 16.5 even if not setting `--reruns`

----

If `pytest-rerunfailures` is installed (version 16.5) then even if not using it via `--reruns` with `pytest`, I'll still get the error: `--reruns incompatible with --pdb` when using `--pdb`. To get around this, I have to either uninstall `pytest-rerunfailures`, or downgrade it to an earlier version if I want to use the `--pdb` functionality.

`pytest SCRIPT.py --pdb` is all you need to reproduce this.

icemac:
Fixes #342.

Since 16.5, `check_options()` runs unconditionally in `pytest_configure` (#332), and its guard `config.option.reruns != 0` is always true when `--reruns` is unset (`None != 0`). Now the check resolves the actually configured reruns (CLI, ini, `--force-reruns`), and marker-driven reruns are rejected per item as in 16.4.

### DIFF
--- a/src/pytest_rerunfailures.py
+++ b/src/pytest_rerunfailures.py
@@ -173,7 +173,6 @@ def _get_global_reruns(config):
     if reruns is not None:
         return reruns
 
-    reruns = None
     with suppress(TypeError, ValueError):
         reruns = int(config.getini("reruns"))
     return reruns
--- a/src/pytest_rerunfailures.py
+++ b/src/pytest_rerunfailures.py
@@ -168,6 +168,17 @@ def pytest_addoption(parser):
     )
 
 
+def _get_global_reruns(config):
+    reruns = config.getvalue("reruns")
+    if reruns is not None:
+        return reruns
+
+    reruns = None
+    with suppress(TypeError, ValueError):
+        reruns = int(config.getini("reruns"))
+    return reruns
+
+
 # making sure the options make sense
 # should run before / at the beginning of pytest_cmdline_main
 def check_options(config):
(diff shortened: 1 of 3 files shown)


src/pytest_rerunfailures.py
```diff
@@ -168,6 +168,17 @@ def pytest_addoption(parser):
     )
 
 
+def _get_global_reruns(config):
+    reruns = config.getvalue("reruns")
+    if reruns is not None:
+        return reruns
+
+    reruns = None
```
sallner: I do not understand this line. The check in L173 should ensure that `reruns is None`.

It seems that `.getvalue()` is deprecated anyhow https://github.com/pytest-dev/pytest/blob/68308aa288e00ff84880572ed9b3590f6cd7d470/src/_pytest/config/__init__.py#L2096 and the recommended `.getoption()` actually allows to set a default value which could be None here. https://github.com/pytest-dev/pytest/blob/68308aa288e00ff84880572ed9b3590f6cd7d470/src/_pytest/config/__init__.py#L2071

sallner (approved):
I generally like the changes and they look sound. Tests were added that cover the mentioned problem.

src/pytest_rerunfailures.py
icemac: Good catch — the `reruns = None` line was indeed redundant (the check above guarantees it). This PR only moved `_get_global_reruns` unchanged from below (it was introduced in #328); GitHub renders the moved block as new lines. Removed in 7fe96c3.

Regarding `.getvalue()`: agreed, it is soft-deprecated in pytest. Since the plugin uses `getvalue` in several more places, I opened #345 to switch them all consistently in a separate cleanup. (Note that `getoption`'s `default` only applies when the option doesn't exist at all, not when its value is `None` — so for our registered options the switch is purely cosmetic.)

— _Comment created by Claude_

PR #348, opened 2026-08-19
Fix only_rerun marker precedence test

janmrow:
`test_only_rerun_flag_in_flaky_marker` builds CLI arguments but doesn't pass them to `runpytest()`. This means its marker-vs-CLI cases never exercise the CLI option.

Pass the arguments to `runpytest()` so the test verifies that marker-level `only_rerun` takes precedence over the CLI value.

To confirm the gap, I temporarily reversed the precedence. The original test still passed, while the corrected test failed as expected.


janmrow:
This is a test-only one-line fix, so I left out a changelog entry in line with the “unless the change is trivial” guidance in `CONTRIBUTING.rst`. I’m happy to add a `changes/348.misc.rst` entry if preferred.

icemac (approved):

PR #351, opened 2026-08-24
Restore higher-scoped teardown when a flaky condition is falsy

teddytennant:
`pytest_runtest_teardown` decides whether to suspend module, class and session finalizers using a gate that checks the rerun limit, the failure statuses and terminal errors. It never checks the `flaky` marker's `condition`, which `_should_not_rerun` does check.

So with a falsy condition on a failing test, teardown suspends the higher-scoped finalizers, the protocol then declines to rerun, `_restore_suspended_finalizers` is never reached, and that fixture's teardown never runs again for the rest of the session:

```python
def module_fixture():
    yield
    print("module teardown")

@pytest.mark.flaky(reruns=2, condition=False)
def test_fail():
    assert False
```

`module teardown` is not printed. With `condition=True`, or with no marker at all, it is. The damage is not local either, since `suspended_finalizers` is a module-level global: run that file alongside another and the first module's fixture stays un-torn-down past the end of its own module.

Adding the missing term to the gate is the whole fix. I also updated the comment above it, which enumerated the terms and was already one short.

Tests cover class, module and session scope, a string condition that evaluates falsy, and the cross-module case. Two are guards rather than reproductions: `condition=True` still reruns and still tears down exactly once, and a plain `--reruns` failure with no marker is unaffected.

Scope, so you do not have to ask. This fixes the `condition` gap only. Two related leaks are pre-existing and untouched. A terminal error raised in the *teardown* phase still leaks, because `_terminal_errors["teardown"]` is written from `pytest_runtest_makereport` after the teardown hook has already suspended, so it needs a hookwrapper rather than another gate term. And a call-phase failure consisting solely of failed subtests tears higher-scoped fixtures down prematurely on every attempt, which is the same drift in the opposite direction. Happy to take either as a follow-up if you want them.

### DIFF
--- a/changes/351.bugfix.rst
+++ b/changes/351.bugfix.rst
@@ -1,0 +1,1 @@
+Restore module, class, and session scoped fixture teardown when a ``flaky`` marker condition is falsy.
--- a/src/pytest_rerunfailures.py
+++ b/src/pytest_rerunfailures.py
@@ -887,12 +887,14 @@ def pytest_runtest_teardown(item, nextitem):
         return
 
     # Only remove non-function level actions from the stack if the test is to be re-run
-    # Exceeding re-run limits, being free of failue statuses, and encountering
-    # allowable exceptions indicate that the test is not to be re-ran.
+    # Exceeding re-run limits, being free of failue statuses, encountering
+    # allowable exceptions, and a falsy flaky condition indicate that the test is
+    # not to be re-ran.
     if (
         item.execution_count <= reruns
         and any(_test_failed_statuses.values())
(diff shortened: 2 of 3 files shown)


icemac (approved):

icemac:
@teddytennant Than you for this PR. If you have time and energy please also fix the other two leaks you detected. I do not think there are even tickets for them already. (But fixing them in a PR is enough, no need to file an issue first.)

teddytennant:
Will do. I'll send them as two separate PRs, starting with the teardown-phase terminal error since that one needs the hookwrapper change.

PR #352, opened 2026-08-24
Fix subtest rerun reporting with xdist

### LINKED ISSUE #350: Successful subtest rerun still reports failure under xdist
This seems to be a follow-up to #315 and #330.

Currently, with pytest built-in subtests, a subtest that fails on the first attempt and passes after a rerun, behaves correctly in a serial run. But still leaves the session failed under xdist:

```python
attempt = 0


def test_subtest_recovers(subtests):
    global attempt
    attempt += 1

    with subtests.test():
        assert attempt >= 2
```

### Without xdist

```console
pytest test_repro.py --reruns 1
```

```text
1 passed, 1 subtests passed, 1 rerun
```

Exit code: `0`.

### With xdist

```console
pytest test_repro.py --reruns 1 -n 1
```

```text
1 failed, 1 passed, 1 subtests passed, 1 rerun
```

Exit code: `1`.

The failed `SubtestReport` from the first attempt remains accounted for on the controller. Even though the rerun succeeds. When the subtest fails on both attempts, the serial run reports two failures (the final failed subtest and its parent test), while xdist reports three: those two and the superseded first-attempt subtest report.

It seems that pytest emits and xdist forwards the `SubtestReport` before pytest-rerunfailures decides to rerun the parent test. The later cleanup does not remove the already forwarded report, or its failure accounting from the controller.

Reproduced on current `master` ([`eab650f`](https://github.com/pytest-dev/pytest-rerunfailures/commit/eab650fb2a2665d61539aafda81ae549440c2834)) with pytest 9.1.1 and pytest-xdist 3.8.0.

If this fits the project's scope, I'd be happy to try to work on a fix.

janmrow:
Fixes #350.

This fixes an issue where a built-in subtest that passes on rerun could still leave a stale failure under xdist, causing the session to fail.

The fix removes only failures from the superseded attempt, while preserving results from other workers and duplicate collected items.

Regression tests cover successful and exhausted reruns, cross-worker isolation, and duplicate collection under xdist.

### DIFF
--- a/changes/350.bugfix.rst
+++ b/changes/350.bugfix.rst
@@ -1,0 +1,1 @@
+Prevent superseded built-in subtest failures from remaining in the final test result when rerunning tests with pytest-xdist.
--- a/src/pytest_rerunfailures.py
+++ b/src/pytest_rerunfailures.py
@@ -410,10 +410,12 @@ def _remove_failed_subtests_from_report(item, report):
         del failed_subtests[report.nodeid]
 
 
-def _remove_failed_subtest_reports_from_stats(item):
+def _remove_failed_subtest_reports_from_stats(
+    config, session, nodeid, worker_id=None, item_index=None
+):
     """
-    Remove already-logged SubtestReports for this item from the terminal reporter's
-    stats buckets.
+    Remove already-logged SubtestReports for a superseded test attempt from the
+    terminal reporter's stats buckets.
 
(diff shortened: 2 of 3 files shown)


icemac (approved):
LGTM but let me first add a test run against xdist so your tests (and some other xdist tests) are actually run in CI.

icemac:
Done in #354.

PR #355, opened 2026-08-27
Fix set-output deprecation warnings

icemac:
Each matrix job warned about the deprecated `set-output` command; write to `$GITHUB_OUTPUT` instead. `matrix.os` is gone from the test cache key because the `tests` job has no `os` dimension.

— _Comment created by Claude_

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -47,16 +47,16 @@ jobs:
       - name: Get pip cache dir
         id: pip-cache
         run: |
-          echo "::set-output name=dir::$(pip cache dir)"
+          echo "dir=$(pip cache dir)" >> "$GITHUB_OUTPUT"
 
       - name: Cache
         uses: actions/cache@v6
         with:
           path: ${{ steps.pip-cache.outputs.dir }}
           key:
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -47,16 +47,16 @@ jobs:
...
       - name: Cache
         uses: actions/cache@v6
         with:
           path: ${{ steps.pip-cache.outputs.dir }}
           key:
-            ${{ matrix.os }}-${{ matrix.python-version }}-${{ matrix.pytest-version }}-v1-${{ hashFiles('**/pyproject.toml') }}
+            ${{ matrix.python-version }}-${{ matrix.pytest-version }}-v1-${{ hashFiles('**/pyproject.toml') }}
           restore-keys: |
-            ${{ matrix.os }}-${{ matrix.python-version }}-${{ matrix.pytest-version }}-v1-
+            ${{ matrix.python-version }}-${{ matrix.pytest-version }}-v1-
```
sallner, line 59: I do not understand those changes. Why do we want to cache all things together? Is it because the eggs/wheels are os-labelled anyhow?

sallner (approved):

.github/workflows/test.yml
icemac: The commit message has: "Also drop matrix.os from the test cache key: the tests job has no os matrix dimension, so it only added an empty segment."

PR #356, opened 2026-08-27
Restore higher-scoped teardown when a teardown error rules out a re-run

teddytennant:
Follow-up to #351, the first of the two leaks I mentioned there.

`pytest_runtest_teardown` takes the module, class and session finalizers off the setup stack when it expects a re-run. Whether the error a teardown raised is terminal is only worked out afterwards, in `pytest_runtest_makereport`, so that decision cannot see it. When `--rerun-except` or `--only-rerun` classifies that error as one not to re-run on, the protocol then declines to re-run, `_restore_suspended_finalizers` is never reached, and the fixture's teardown never runs again for the rest of the session:

```python
def module_fixture():
    yield
    print("module teardown")

@pytest.fixture
def broken_fixture():
    yield
    raise ValueError("teardown error")

@pytest.mark.flaky(reruns=2, rerun_except=["ValueError"])
def test_fail(broken_fixture):
    assert False
```

`module teardown` is not printed. As with #351 the damage is not local, since `suspended_finalizers` is a module-level global: run that file alongside another and the first module's fixture stays un-torn-down past the end of its own module.

Adding a term to the gate does not work here, because the error does not exist yet when the gate runs. So the hook becomes a wrapper. On the way in it decides exactly as before. On the way out it looks at the error the teardown phase actually raised, and if that error rules out a re-run it puts the finalizers back and calls `teardown_exact`. Running them there rather than just restoring them matters: a restore alone leaves a stale collector on the stack, and the next item's `SetupState.setup` asserts on that.

If a higher-scoped teardown then fails itself, its error is chained onto the one that got us there instead of replacing it. Replacing it would change what the re-run decision is made on, and the test would start re-running again.

Tests cover class, module and session scope, the cross-module case, and a higher-scoped teardown that fails while being run this way. Two are guards rather than reproductions: an error the filters do allow a re-run on still re-runs and still tears the module fixture down exactly once, and the plain `--reruns` paths are untouched.

The remaining leak, premature teardown when a call phase fails through subtests only, is a separate PR.

### DIFF
--- a/src/pytest_rerunfailures.py
+++ b/src/pytest_rerunfailures.py
@@ -13,9 +13,12 @@ from contextlib import suppress
 from typing import Any
 
 import pytest
-from _pytest.outcomes import fail
+from _pytest.outcomes import Exit, fail
 from _pytest.runner import runtestprotocol
 from packaging.version import parse as parse_version
+
+if sys.version_info < (3, 11):
+    from exceptiongroup import BaseExceptionGroup
 
 failed_subtests_key: Any
 SubtestReport: Any
@@ -534,10 +537,6 @@ def _should_hard_fail_on_error(item, report, excinfo):
     if report.outcome != "failed":
         return False
 
-    return _is_terminal_error(item, excinfo)
-
-
-def _is_terminal_error(item, excinfo):
     rerun_errors = _get_rerun_filter_regex(item, "only_rerun")
     rerun_except_errors = _get_rerun_filter_regex(item, "rerun_except")
 
@@ -902,21 +901,65 @@ def _is_rerun_path_excluded(item):
     )
 
 
-def _suspend_finalizers_for_rerun(item):
-    """Hold back the higher-scope finalizers if the test is going to be re-run.
+def _teardown_suspended_finalizers(item, call, report):
+    """Tear down the scopes held back for a re-run that will not happen.
 
-    Returns whether they were taken off the stack.
+    pytest_runtest_teardown takes the module, class and session finalizers off
+    the setup stack when it expects the test to be re-run. Whether the error a
+    teardown raised rules that re-run out is only known here, where the report
+    of the phase says how the error was classified, so put the finalizers back
+    and tear down whatever this item was the last user of.
+
+    Returns the report of the teardown phase.
     """
+    if not getattr(item, "_finalizers_suspended", False):
+        return report
+
+    item._finalizers_suspended = False
+    _restore_suspended_finalizers(item)
+
+    def teardown_higher_scopes():
+        try:
+            item.session._setupstate.teardown_exact(item._teardown_nextitem)
+        except (Exit, KeyboardInterrupt):
+            # A fixture that ends the session has to keep ending it.
+            raise
+        except BaseException as exc:
+            if call.excinfo is None:
+                raise
+            # Both errors come from the teardown of this item, so report them
+            # together, the way pytest reports several failing finalizers.
+            raise BaseExceptionGroup(
+                "errors during test teardown", [exc, call.excinfo.value]
+            ) from None
+
+    teardown_call = pytest.CallInfo.from_call(
+        teardown_higher_scopes, when="teardown", reraise=(Exit, KeyboardInterrupt)
+    )
+    if teardown_call.excinfo is None:
+        return report
+
+    # The report of the phase is already built, so the only way for the error
+    # above to reach the terminal is a report that replaces it.
+    return pytest.TestReport.from_item_and_call(item, teardown_call)
+
+
+def pytest_runtest_teardown(item, nextitem):
+    # pytest_runtest_makereport needs both of these to finish a teardown this
+    # hook left half done because it expected a re-run.
+    item._finalizers_suspended = False
+    item._teardown_nextitem = nextitem
+
     reruns = get_reruns_count(item)
     if reruns is None:
         # global setting is not specified, and this test is not marked with
         # flaky
-        return False
+        return
 
     if not hasattr(item, "execution_count"):
         # pytest_runtest_protocol hook of this plugin was not executed
         # -> teardown needs to be skipped as well
-        return False
+        return
 
     _test_failed_statuses = getattr(item, "_test_failed_statuses", {})
 
@@ -926,7 +969,7 @@ def _suspend_finalizers_for_rerun(item):
         and item.session.config.failures_db.get_suite_reruns() >= max_suite_reruns
     ):
         _restore_suspended_finalizers(item)
-        return False
+        return
 
     # Only remove non-function level actions from the stack if the test is to be re-run
     # Exceeding re-run limits, being free of failue statuses, encountering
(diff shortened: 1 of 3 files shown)


src/pytest_rerunfailures.py
```diff
@@ -940,9 +948,41 @@ def pytest_runtest_teardown(item, nextitem):
...
+    return False
+
+
+def _teardown_error_is_terminal(item, outcome):
+    """Report whether the teardown phase raised an error that stops re-runs."""
+    exc_info = outcome.excinfo
+    if exc_info is None:
+        return False
+
+    return _is_terminal_error(item, pytest.ExceptionInfo.from_exc_info(exc_info))
```
icemac: **This drops the `report.outcome != "failed"` guard, so non-failing teardown exceptions trigger a premature teardown + re-setup on every rerun.**

`_should_hard_fail_on_error` only consults `_is_terminal_error` when `report.outcome == "failed"`. This wrapper calls `_is_terminal_error` on the raw exception with no such guard, so the two disagree: a teardown exception that pytest does *not* report as a failure is terminal here, while `pytest_runtest_makereport` sets `_terminal_errors["teardown"] = False` and the protocol still reruns the test — after this hook has already restored and run the higher-scoped finalizers.

```python
def module_fixture():
    print("MODULE SETUP"); yield; print("MODULE TEARDOWN")

@pytest.fixture
def skipping_teardown():
    yield
    pytest.skip("skip in teardown")

@pytest.mark.flaky(reruns=2, only_rerun=["AssertionError"])
def test_fail(skipping_teardown):
    assert False
```

`master`: `MODULE SETUP` once, `MODULE TEARDOWN` once. This branch: setup and teardown three times — the exact leak class this series removes.

`Exit` and `KeyboardInterrupt` diverge the same way: pytest re-raises them out of the teardown hook (`CallInfo.from_call(reraise=(Exit, KeyboardInterrupt))`) and never turns them into a failed report.

— _Comment created by Claude_

src/pytest_rerunfailures.py
```diff
@@ -940,9 +948,41 @@ def pytest_runtest_teardown(item, nextitem):
...
+    # The error a teardown raises is only classified once this hook has
+    # returned, so the decision above could not take it into account. A
+    # terminal one means there is no re-run to hold the finalizers back for,
+    # so put them back on the stack and tear down what this item was the last
+    # user of.
+    if suspended and _teardown_error_is_terminal(item, outcome):
         _restore_suspended_finalizers(item)
+        try:
+            item.session._setupstate.teardown_exact(nextitem)
+        except BaseException as exc:
```
icemac: **`except BaseException` swallows `Exit` and `KeyboardInterrupt` raised by the higher-scoped finalizers.**

A module fixture whose teardown calls `pytest.exit("EXIT FROM TEARDOWN")`, reached through this path, is silently ignored: the session continues and the next module's test runs (`1 failed, 1 passed, 1 error`). Ctrl-C during that teardown is discarded the same way.

Catching `Exception`, or re-raising `Exit`/`KeyboardInterrupt` before the chaining, avoids this.

— _Comment created by Claude_

src/pytest_rerunfailures.py
```diff
@@ -940,9 +948,41 @@ def pytest_runtest_teardown(item, nextitem):
...
+            # The error raised above is the one that rules out a re-run, so it
+            # has to stay the error of this phase. Chain this one onto it
+            # instead of replacing it, so both end up in the report.
+            outcome.excinfo[1].__context__ = exc
```
icemac: **Attaching the secondary error via `__context__` loses it whenever the primary exception suppresses its context.**

Assigning `__context__` does not clear `__suppress_context__` or `__cause__`, so if the original teardown exception was raised with `from None` (or `from something`) the chained error never reaches the report. With `raise ValueError("teardown error") from None` in the function-scoped fixture, the module fixture's `RuntimeError("MODULE TEARDOWN ERROR")` vanishes from the output entirely. It also clobbers any pre-existing `__context__` on the primary exception.

Minor, related: because the later error is installed as the *context* of the earlier one, the chained report reads backwards — `RuntimeError: module teardown error`, then `During handling of the above exception, another exception occurred:`, then the `ValueError` that actually happened first.

— _Comment created by Claude_

src/pytest_rerunfailures.py
```diff
@@ -940,9 +948,41 @@ def pytest_runtest_teardown(item, nextitem):
                     if key not in suspended_finalizers:
                         suspended_finalizers[key] = item.session._setupstate.stack[key]
                     del item.session._setupstate.stack[key]
-    else:
-        # restore suspended finalizers
+        return True
```
icemac: **Minor: this `return True` sits outside the `if item in item.session._setupstate.stack:` block**, so a `True` return does not imply the docstring's "they were taken off the stack".

When the item is not on the stack, the caller still treats the finalizers as suspended and, on a terminal teardown error, runs `_restore_suspended_finalizers` (splicing in whatever stale globals are left from a previous item) plus a redundant `teardown_exact`.

I could not construct a case where this actually misbehaves today — the gate's `not any(item._terminal_errors.values())` rules out the collector-setup-failure paths — so this is a contract mismatch rather than a live bug.

— _Comment created by Claude_

icemac (commented):
Reviewed against real repro cases (pytest 9.1.1) and the full suite with and without xdist + subtests (179 passed).

The first comment below is a regression against this PR's own goal, so it is the one that matters; the rest are on the new error-chaining path.

Also checked and clean: module/class fixture preservation when the next item shares the scope, terminal teardown error on the second attempt, and stack ordering after `stack.update(suspended_finalizers)`. `test_run_session_teardown_once_after_reruns` fails under pytest 8.2, but identically on `master` — pre-existing, not from this PR.

— _Comment created by Claude_

src/pytest_rerunfailures.py
teddytennant: Confirmed, and `pytest.xfail()` in a teardown goes the same way, since `skipping.py` turns that into a skipped report too. I could not write a guard here I trusted: the wrapper runs before the report exists, so any list of exception types in it is a copy of pytest's classification that will drift.

So the restore and the `teardown_exact` moved into `pytest_runtest_makereport`, where the report and `_terminal_errors` already exist, and `_is_terminal_error` is gone. One derivation instead of two, and the two cannot disagree by construction. Your repro and the xfail one both tear the module fixture down once again, same as master.

src/pytest_rerunfailures.py
teddytennant: Fixed. The extra teardown runs inside `CallInfo.from_call` with the same `reraise=(Exit, KeyboardInterrupt)` pytest uses, so both get out and stop the session. For the record `except Exception` on its own would not have covered it either, `Exit` subclasses `Exception`.

src/pytest_rerunfailures.py
teddytennant: Good catch, the `__context__` assignment is gone. When both teardowns fail the two errors go into a `BaseExceptionGroup("errors during test teardown")` now, which is what pytest itself builds for several failing finalizers, so `from None` is no longer a factor and the order reads the way pytest's does.

src/pytest_rerunfailures.py
teddytennant: Tightened. The flag is set on the line that deletes from the stack now, so it is true only if something actually came off. Still no test for it, since I could not reach that state either.

teddytennant:
Pushed as a new commit. It reads smaller than it sounds: `pytest_runtest_teardown` is back to master's plain hookimpl plus three lines recording what it held back, and `pytest_runtest_makereport` does the work once the real report exists.

The exception group needs `from exceptiongroup import BaseExceptionGroup` on 3.10. pytest already requires `exceptiongroup` there and `_pytest/runner.py` imports it the same way, so I left it transitive rather than touching `pyproject.toml` in a bugfix. Say the word if you would rather have it declared.

183 passed on 3.13 with pytest 9.1.1, same under `-n 2`, and 175 passed 8 skipped on 3.10 with pytest 8.2.1. Four new tests, all of which fail on the previous commit.

icemac (approved):

PR #357, opened 2026-08-27
Keep higher-scoped fixtures alive across re-runs of failed subtests

teddytennant:
Follow-up to #351, the second of the two leaks I mentioned there. Independent of #356, though both touch the same condition in `pytest_runtest_teardown`, so whichever lands second wants a one line rebase. Happy to do it.

A call phase that fails through subtests only leaves the call report itself passing, so every entry in `_test_failed_statuses` is false. `pytest_runtest_teardown` reads that as a finished test and lets the module, class and session finalizers run, while `_should_not_rerun` does count the failed subtests and re-runs the test anyway. The higher-scoped fixtures are torn down and set up again on every attempt:

```python
def module_fixture():
    print("module setup")
    yield
    print("module teardown")

attempts = []

def test_subtests(subtests):
    attempts.append(1)
    with subtests.test("fails twice"):
        assert len(attempts) > 2
```

With `--reruns 3` that prints three setups and three teardowns. The same test failing normally, without subtests, prints one of each.

The fix counts the failed subtests in the teardown check, the same way the re-run decision already does. `_get_num_failed_subtests` takes a nodeid now instead of a report, since the teardown hook has no report to hand it.

Tests cover class, module and session scope and the cross-module case. One is a guard rather than a reproduction: subtests that never pass exhaust the re-runs and still tear the module fixture down exactly once.

### DIFF
--- a/src/pytest_rerunfailures.py
+++ b/src/pytest_rerunfailures.py
@@ -484,7 +484,7 @@ def _remove_failed_subtest_reports_from_stats(
     _remove_subtest_reports("subtests passed")
 
 
-def _get_num_failed_subtests(item, report):
+def _get_num_failed_subtests(item, nodeid):
     """
     Return the number of failed subtests.
 
@@ -495,7 +495,7 @@ def _get_num_failed_subtests(item, report):
 
     failed_subtests = item.config.stash.get(failed_subtests_key, None)
     if failed_subtests is not None:
-        return failed_subtests.get(report.nodeid, 0)
+        return failed_subtests.get(nodeid, 0)
 
     return 0
 
@@ -566,7 +566,7 @@ def _should_not_rerun(item, report, reruns):
     is_terminal_error = any(item._terminal_errors.values())
     condition = get_reruns_condition(item)
     has_failed_subtests = (
-        report.when == "call" and _get_num_failed_subtests(item, report) > 0
+        report.when == "call" and _get_num_failed_subtests(item, report.nodeid) > 0
     )
 
     return (
@@ -974,10 +974,15 @@ def pytest_runtest_teardown(item, nextitem):
     # Only remove non-function level actions from the stack if the test is to be re-run
     # Exceeding re-run limits, being free of failue statuses, encountering
     # allowable exceptions, and a falsy flaky condition indicate that the test is
-    # not to be re-ran.
+    # not to be re-ran. A failure can also be carried by failed subtests alone,
+    # which leaves the call phase itself passing.
     if (
         item.execution_count <= reruns
-        and any(_test_failed_statuses.values())
+        and (
+            any(_test_failed_statuses.values())
+            or _get_num_failed_subtests(item, item.nodeid) > 0
+        )
+        and not any(item._test_xfailed.values())
         and not any(item._terminal_errors.values())
         and get_reruns_condition(item)
     ):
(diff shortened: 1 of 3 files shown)


src/pytest_rerunfailures.py
```diff
@@ -923,10 +923,14 @@ def pytest_runtest_teardown(item, nextitem):
...
     # allowable exceptions, and a falsy flaky condition indicate that the test is
-    # not to be re-ran.
+    # not to be re-ran. A failure can also be carried by failed subtests alone,
+    # which leaves the call phase itself passing.
     if (
         item.execution_count <= reruns
-        and any(_test_failed_statuses.values())
+        and (
+            any(_test_failed_statuses.values())
+            or _get_num_failed_subtests(item, item.nodeid) > 0
```
icemac, line 1009: **This condition no longer mirrors `_should_not_rerun`: the `xfail` branch has no counterpart, so higher-scope finalizers get suspended and never restored.**

`_should_not_rerun` returns `True` — no re-run — when `hasattr(report, "wasxfail")`, regardless of failed subtests. Before this PR the two stayed in sync by accident: an xfailed/xpassed call report leaves every entry in `_test_failed_statuses` false, so the `else` branch below ran `_restore_suspended_finalizers`. Adding the `_get_num_failed_subtests(...) > 0` disjunct breaks that — teardown now strips the module/class/session finalizers off `session._setupstate.stack` into the global `suspended_finalizers`, the protocol then declines to re-run because of `wasxfail`, and nothing ever puts them back.

Repro (subtest fails, then the test xfails itself):

```python
# test_a.py
def module_fixture():
    print("MODULE A SETUP"); yield; print("MODULE A TEARDOWN")

def test_xfail_subtest(subtests):
    with subtests.test("fails"):
        assert False
    pytest.xfail("known issue")
```

`pytest -s --reruns 2`:
- `master`: `MODULE A SETUP` … `MODULE A TEARDOWN`
- this branch: `MODULE A SETUP` only — the module finalizer never runs

With a second module present it is worse than a skipped finalizer: the next item's `stack.update(...)` re-inserts it *after* module B's node, so module A is torn down after module B's test has already run — the ordering corruption #351 fixed.

Suggested fix: record the xfail state alongside the failure statuses (in `pytest_runtest_makereport`, `item._test_xfailed[result.when] = hasattr(result, "wasxfail")`) and add `and not any(item._test_xfailed.values())` here, so the condition mirrors every disjunct of `_should_not_rerun` rather than four of the five.

— _Comment created by Claude_

icemac (commented):
Reviewed with a full diff pass plus runtime verification against pytest 9.1.1. The change does what it targets — subtest-only failures now keep module/class/session fixtures alive, and the `pytest.skip()`-after-failed-subtest path still re-runs without leaking. All 178 existing tests plus the 5 new ones pass, and the new tests themselves look right (module ordering is alphabetically stable; `temporary_failure`'s `test.res` marker is per-tmpdir, so the `scope` parametrization does not cross-contaminate).

One regression, in the comment below.

Two things I looked at and am *not* filing:

- `--only-rerun` / `--rerun-except` are bypassed for subtest-only failures, because `_should_hard_fail_on_error` returns early when the call report's outcome is `passed`. Teardown and the re-run decision agree there, so no finalizer leak — pre-existing behaviour from the subtests support, not this PR.
- The `failed_subtests` stash entry is never deleted on the final, non-re-run attempt. Harmless today: teardown only reads the current item's nodeid, and each nodeid runs one protocol per session.

— _Comment created by Claude_

src/pytest_rerunfailures.py
teddytennant: Fixed, and rebased onto master now that #356 landed.

Reproduced both symptoms first: the module finalizer never running, and with a second module present, `MODULE A TEARDOWN` landing after `TEST B RUNS`.

One change from your suggestion. Plain assignment works, but only because of report ordering. pytest-subtests emits its own `call` report, so two reports collide on that key:

```
when=call wasxfail=False   <- subtest report
when=call wasxfail=True    <- the real call report
```

The xfail one lands last today. That's the same ordering dependency in `_test_failed_statuses` that this PR is working around, so I made it accumulate rather than overwrite.

New test is `test_xfail_after_failing_subtest_restores_module_fixture`. It uses two modules so it catches the ordering corruption and not just the missing finalizer, and it fails without the new disjunct. 183 passed, 6 skipped.

Agreed the other two are out of scope here.

icemac (approved):

PR #358, opened 2026-08-27
Authenticate xdist StatusDB connections

janmrow:
Generate a per-session token on the xdist controller and pass it to workers through `workerinput`.

StatusDB now requires a successful handshake before processing commands and rejects clients with an invalid token.

Tests cover the authentication check and passing the connection details to workers.

### DIFF
--- a/src/pytest_rerunfailures.py
+++ b/src/pytest_rerunfailures.py
@@ -757,6 +757,8 @@ class SocketDB(StatusDB):
         buf = b""
         while True:
             b = conn.recv(1)
+            if not b:
+                raise ConnectionError("StatusDB connection closed unexpectedly")
             if b == self.delim:
                 break
             buf += b
--- a/src/pytest_rerunfailures.py
+++ b/src/pytest_rerunfailures.py
@@ -753,7 +753,7 @@ class SocketDB(StatusDB):
         self.sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
         self.sock.setblocking(1)
 
-    def _sock_recv(self, conn) -> str:
+    def _sock_recv_bytes(self, conn) -> bytes:
         buf = b""
         while True:
             b = conn.recv(1)
@@ -763,7 +763,10 @@ class SocketDB(StatusDB):
                 break
             buf += b
 
-        return buf.decode()
+        return buf
+
+    def _sock_recv(self, conn) -> str:
+        return self._sock_recv_bytes(conn).decode()
 
     def _sock_send(self, conn, msg: str):
         conn.send(msg.encode() + self.delim)
@@ -793,7 +796,9 @@ class ServerStatusDB(SocketDB):
 
     def run_connection(self, conn):
         with conn, suppress(ConnectionError):
-            authenticated = secrets.compare_digest(self._sock_recv(conn), self.token)
+            authenticated = secrets.compare_digest(
+                self._sock_recv_bytes(conn), self.token.encode("ascii")
+            )
             self._sock_send(conn, "1" if authenticated else "0")
             if not authenticated:
                 return
--- a/src/pytest_rerunfailures.py
+++ b/src/pytest_rerunfailures.py
@@ -753,7 +753,7 @@ class SocketDB(StatusDB):
         self.sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
         self.sock.setblocking(1)
 
-    def _sock_recv_bytes(self, conn) -> bytes:
+    def _sock_recv_bytes(self, conn, max_length: int | None = None) -> bytes:
         buf = b""
         while True:
             b = conn.recv(1)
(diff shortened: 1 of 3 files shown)


src/pytest_rerunfailures.py
```diff
@@ -781,7 +787,12 @@ def run_server(self):
             t.start()
 
     def run_connection(self, conn):
-        with suppress(ConnectionError):
+        with conn, suppress(ConnectionError):
+            authenticated = secrets.compare_digest(self._sock_recv(conn), self.token)
```
icemac: **Non-ASCII or invalid-UTF-8 handshake input kills this thread instead of rejecting the client.**

`secrets.compare_digest` raises `TypeError` when either `str` argument contains non-ASCII characters, and `_sock_recv`'s `buf.decode()` raises `UnicodeDecodeError` on invalid UTF-8. Neither is a `ConnectionError`, so `suppress(ConnectionError)` does not catch them.

Sending `"\u00e9\n"` to the port on this branch dumps

```
Exception in thread Thread-3 (run_connection):
  ...
  authenticated = secrets.compare_digest(self._sock_recv(conn), self.token)
TypeError: comparing strings with non-ASCII characters is not supported
```

to stderr in the middle of the user's test run, and the client receives `b''` rather than the intended `"0"` rejection.

Since this handshake is exactly the code now expected to face arbitrary local input, it should compare raw bytes (recv the token undecoded and `compare_digest(raw, self.token.encode())`), or wrap the comparison so any malformed input yields a clean `"0"`.

— _Comment created by Claude_

icemac (commented):
Reviewed the handshake against the branch (suite also run under xdist: 175 passed). The change itself is sound — the token is assigned before `run_server` starts, the `with conn, suppress(...)` nesting order is right, `failures_db` is always a `ServerStatusDB` where `.token` is read, the token survives execnet serialization, and both new tests fail before the change and pass after.

Three things on the error paths, though; the first two chain into a session hang under xdist.

— _Comment created by Claude_

src/pytest_rerunfailures.py
```diff
@@ -848,9 +859,13 @@ def get_suite_reruns(self) -> int:
 
 
 class ClientStatusDB(SocketDB):
-    def __init__(self, sock_port):
+    def __init__(self, sock_port, token):
         super().__init__()
         self.sock.connect(("127.0.0.1", sock_port))
+        self._sock_send(self.sock, token)
+        if self._sock_recv(self.sock) != "1":
```
icemac, line 880: **A worker can hang forever here (busy-spinning) if the server never answers the handshake.**

`_sock_recv` loops on `conn.recv(1)` and only breaks on the `b"\n"` delimiter. At EOF `recv` returns `b""`, which is neither the delimiter nor an error, so the loop never terminates.

That is now directly reachable: whenever the server-side `run_connection` thread dies before replying (see the `compare_digest` comment above, or any other non-`ConnectionError` exception), the new `with conn` closes the socket, and this call sits in `_sock_recv` burning 100% of a core — the `ConnectionError("StatusDB authentication failed")` below is never raised. The xdist controller then waits on a worker that never finishes `pytest_configure`, hanging the whole session.

The same EOF gap leaves a live spinning thread on the server side: connecting and immediately closing takes `threading.active_count()` from 2 to 3 permanently.

`_sock_recv` should treat `b""` as end-of-stream and raise `ConnectionError`; a `settimeout` on the handshake would be worth considering too.

— _Comment created by Claude_

src/pytest_rerunfailures.py
```diff
@@ -781,7 +787,12 @@ def run_server(self):
             t.start()
```
icemac, line 797: **Minor: the handshake buffers unboundedly before authentication, and connections are unlimited.**

`_sock_recv` appends to `buf` with no length cap, and every accepted connection gets its own daemon thread with no connection limit — both before any authentication has happened. Any local process can connect and stream bytes without a newline until the controller runs out of memory.

Only a local DoS, but capping the pre-auth read at the known token length (64 hex chars) is nearly free and fits the threat model this PR is addressing.

— _Comment created by Claude_

janmrow: Fixed in 081a2e6. The pre-authentication read is now limited to the expected token length, while normal StatusDB messages remain unchanged.

src/pytest_rerunfailures.py
janmrow: Fixed in f9f0be6. `_sock_recv()` now treats EOF as a closed connection and raises `ConnectionError` instead of continuing to read.

I considered adding a handshake timeout too but left it out for now since that covers the broader connection case and I guess it would introduce a timeout policy. Happy to add it if you think it belongs in this PR.

src/pytest_rerunfailures.py
janmrow: Fixed in f72bc29. The authentication token is now read and compared as raw bytes, so invalid UTF-8 and non-ASCII input are rejected.

icemac (approved):

PR #360, opened 2026-09-02
Prevent negative reruns from skipping initial test execution

### LINKED ISSUE #359: Negative rerun counts can skip tests with exit code 0
It looks like a negative rerun count can cause a collected test not to run, while pytest exits successfully.

For example, running `pytest --reruns=-1` with:
```python
def test_should_run():
    assert False
```
collects one test, but does not report a test result and exits with code 0.

The same happens with negative values from:
* `reruns = -1`
* `@pytest.mark.flaky(reruns=-1)`
* `--force-reruns=-1`

`--reruns=0` works as expected here.

Negative rerun counts should probably be rejected as invalid configuration instead of allowing the test session to succeed without running the test.
If this looks worth fixing, I'd be happy to work on it.

janmrow:
Fixes #359.

Negative rerun counts could skip the test entirely, because the rerun limit was checked before the first run.

This change ensures that the initial test execution is never suppressed by the rerun count.
Negative values no longer produce a false success without running the test.

Regression tests cover the supported rerun configuration sources, and also the distributed StatusDB path with xdist.
I also ran some local A/B checks for non-negative rerun counts, xdist worker crashes, exclusions, and other negative values.

### DIFF
--- a/changes/359.bugfix.rst
+++ b/changes/359.bugfix.rst
@@ -1,0 +1,1 @@
+Ensure negative rerun counts do not skip the initial test execution.
--- a/src/pytest_rerunfailures.py
+++ b/src/pytest_rerunfailures.py
@@ -1089,7 +1089,8 @@ def pytest_runtest_protocol(item, nextitem):
     item.execution_count = db.get_test_failures(item.nodeid)
     db.set_test_reruns(item.nodeid, reruns)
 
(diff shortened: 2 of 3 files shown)


icemac (approved):

icemac:
Thank you for this PR. 😃

PR #362, opened 2026-09-06
Fix bare exception class in only_rerun / rerun_except causing INTERNALERROR

janmrow:
Allow `only_rerun` and `rerun_except` in `@pytest.mark.flaky` to accept a single exception class directly.

Before the fix passing a bare exception class name instead of a list, caused an `INTERNALERROR`. The value is now handled as a single filter, consistently with string arguments.

Reproduction:
```python
@pytest.mark.flaky(reruns=2, only_rerun=ValueError)
def test_x():
    raise ValueError("ERR")
```

Before: `INTERNALERROR: TypeError: 'type' object is not iterable`.
After: `1 failed, 2 rerun`.


janmrow:
The `pytest-main` failures here looks that are not related to scope of this PR. As far as I investigated, they are caused by the pytest `NodeId` change. Which is already implemented in #363.

icemac (approved):

icemac:
Thank you for this PR. 😃

PR #363, opened 2026-09-06
Fix subtest reruns with structured NodeId keys

janmrow:
[pytest-dev/pytest#14758](https://github.com/pytest-dev/pytest/pull/14758) changed how pytest stores node IDs internally: instead of strings, it now uses `NodeId` objects.

Because of this, the private `failed_subtests` mapping uses `NodeId` objects as keys.

`pytest-rerunfailures` was still looking up failed subtests using string `.nodeid` values. So that failed subtests were not rerun with newer pytest versions.

This change uses `.id` when `NodeId` is available, and falls back to `.nodeid` for older pytest versions.

Tested locally:
- the full test suite passes with pytest 9.1.1 and pytest main at `431f3e1f5`,
- subtest regression tests pass,
- related xdist/subtest tests pass.

Only the `failed_subtests` handling was changed. Other uses of `.nodeid` stay the same.


icemac (approved):

icemac:
Thank you for this PR. 😃

PR #364, opened 2026-09-06
Match only_rerun against wrapped exception causes

### LINKED ISSUE #353: only_rerun does not match exceptions wrapped by another exception (__cause__ / __context__ not inspected)
### Summary

`--only-rerun` / `only_rerun` matches against the outermost exception only. When a transient error is wrapped by another exception via `raise ... from error`, the pattern can no longer match it, so the test is not retried even though the underlying error is exactly the kind users configure `only_rerun` for.

### Current behaviour

`_try_match_error` builds the string to match from `excinfo` alone (`src/pytest_rerunfailures.py`, master):

```python
def _try_match_error(rerun_errors, excinfo):
    if excinfo:
        err = f"{excinfo.type.__name__}: {excinfo.value}"
        for rerun_error in rerun_errors:
            if isinstance(rerun_error, type) and issubclass(rerun_error, BaseException):
                if issubclass(excinfo.type, rerun_error):
                    return True
            elif re.search(rerun_error, err):
                return True
```

`excinfo.type` and `excinfo.value` are the outermost exception. Neither `__cause__` nor `__context__` is consulted, so both the regex form and the exception-type form miss a wrapped error.

### Reproduction

```python
import pytest


def test_wrapped():
    try:
        raise MemoryError("out of memory")
    except MemoryError as error:
        raise RuntimeError("something failed, see the exception above") from error
```

```
$ pytest test_wrapped.py --reruns 2 --only-rerun MemoryError -rR
1 failed in 0.01s          # not rerun
```

Removing the wrapper makes it retry as expected:

```
$ pytest test_direct.py --reruns 2 --only-rerun MemoryError -rR
1 failed, 2 rerun in 0.02s
```

### Why this matters in practice

We hit this in the TRL CI, where the suite runs under `pytest-xdist` on a shared GPU and `--only-rerun` includes `OutOfMemoryError` to absorb memory pressure between workers.

When the OOM is raised inside `torch.testing.assert_close`, `torch.testing` wraps any exception it does not expect:

```python
raise RuntimeError(
    f"Comparing\n\n"
    f"{pair}\n\n"
    f"resulted in the unexpected exception above. "
    ...
) from error
```

The failure then surfaces as `RuntimeError: Comparing ...`, with `torch.OutOfMemoryError` demoted to `__cause__`. The wrapper message does not carry the original error text, so no `only_rerun` value can match it: neither `OutOfMemoryError` (wrong outer type) nor `CUDA out of memory` (absent from the outer message).

The only workaround available to us is matching the wrapper text itself, which we rejected: `torch.testing` emits that message for *any* unexpected exception during a comparison, so it would also retry genuine defects. A deterministic bug would still fail all attempts, but a nondeterministic one would be silently retried into green, which is precisely what a narrow `only_rerun` is meant to prevent.

### Proposed behaviour

Walk the exception chain and match if any linked exception matches, something like:

```python
def _iter_exception_chain(exc):
    seen = set()
    while exc is not None and id(exc) not in seen:
        seen.add(id(exc))
        yield exc
        exc = exc.__cause__ or exc.__context__


def _try_match_error(rerun_errors, excinfo):
    if not excinfo:
        return False
    for exc in _iter_exception_chain(excinfo.value):
        err = f"{type(exc).__name__}: {exc}"
        for rerun_error in rerun_errors:
            if isinstance(rerun_error, type) and issubclass(rerun_error, BaseException):
                if isinstance(exc, rerun_error):
                    return True
            elif re.search(rerun_error, err):
                return True
    return False
```

With that, the existing `OutOfMemoryError` pattern matches the wrapped case on its own, with no ambiguity and no wrapper-text heuristic.

Two notes on scope:

- This is unrelated to #230. That asked for a user-supplied `condition` callback, whose implementation (#299) was reverted in #304 for `pytest-xdist` incompatibility. Chain walking needs no user callback and no new serialization: it reads type names and `str()` of already-available exceptions, and the decision is still taken worker-side in `pytest_runtest_makereport`, exactly where it is taken today.
- `__cause__` is the unambiguous signal (explicit `raise ... from`). Following `__context__` as well also covers exceptions raised inside an `except` block without `from`, which is the more common accidental wrapping. If following `__context__` is considered too broad, restricting to `__cause__` alone would already fix the case above and could be worth gating behind an option.

Happy to open a PR if the approach looks right.

### Environment

- `pytest-rerunfailures` 16.6 (behaviour unchanged on master), also reproduced on 15.1

teddytennant:
Fixes #353.

`--only-rerun` / `only_rerun` (and `--rerun-except`) match the outermost exception only. A transient error wrapped with `raise ... from` is not retried, even when the inner type is exactly what the filter is for.

```python
def test_wrapped():
    try:
        raise MemoryError("out of memory")
    except MemoryError as error:
        raise RuntimeError("something failed") from error
```

```
$ pytest test_wrapped.py --reruns 2 --only-rerun MemoryError
1 failed            # not rerun
```

The same test without the wrapper reruns as expected. `_try_match_error` built the match string from `excinfo.type` / `excinfo.value` and never walked `__cause__` or `__context__`.

The matcher now walks that chain (skipping a context suppressed by `raise ... from None`) and treats a hit on any linked exception as a match. Existing outer-exception matches are unchanged.

Regression tests cover a `__cause__` wrap, an implicit `__context__` wrap, the exception-class form of `only_rerun`, `rerun_except` on a wrapped error, a suppressed context, and a non-matching control.

```
$ pytest tests/test_pytest_rerunfailures.py -k wrapped
7 passed
$ pytest tests/
201 passed, 7 skipped
```


src/pytest_rerunfailures.py
```diff
@@ -545,12 +545,27 @@ def _matches_any_rerun_except_error(rerun_except_errors, excinfo):
...
+        else:
+            exc = None
+
+
 def _try_match_error(rerun_errors, excinfo):
-    if excinfo:
-        err = f"{excinfo.type.__name__}: {excinfo.value}"
+    if not excinfo:
+        return False
+    for exc in _iter_exception_chain(excinfo.value):
```
icemac: **Chain walking silently disables reruns for `--rerun-except`.**

`_matches_any_rerun_except_error` calls this same helper, so a match *anywhere* in the chain now means "do not rerun". Any exception raised inside an `except`/`finally` block carries an implicit `__context__`, so this fires very often.

```python
def test_a():
    try:
        assert False, "genuine failure"
    except AssertionError:
        raise ConnectionError("network blip")
```

`pytest --reruns 2 --rerun-except AssertionError`
* master: `1 failed, 2 rerun`
* this PR: `1 failed` (0 reruns)

The user excluded `AssertionError` and got a `ConnectionError` excluded instead. There is no opt-out, and the only `rerun_except` test added (`test_rerun_except_matches_wrapped_cause`) asserts the new behavior rather than probing this case.

Suggestion: walk the chain for `only_rerun` only, or restrict the `rerun_except` walk to explicit `raise ... from` (`__cause__`) and skip implicit `__context__`.

— _Comment created by Claude_

src/pytest_rerunfailures.py
```diff
@@ -545,12 +545,27 @@ def _matches_any_rerun_except_error(rerun_except_errors, excinfo):
...
+        err = f"{type(exc).__name__}: {exc}"
         for rerun_error in rerun_errors:
             if isinstance(rerun_error, type) and issubclass(rerun_error, BaseException):
-                if issubclass(excinfo.type, rerun_error):
+                if isinstance(exc, rerun_error):
```
icemac, line 626: **Combining `--only-rerun` and `--rerun-except` now cancels out.** (Real location: `_should_hard_fail_on_error`, currently line 639.)

That function returns `(not matches_rerun_only) or matches_rerun_except`. With both filters widened to the chain, one failure can match both sides. Using the same test as the previous comment:

`pytest --reruns 2 --only-rerun ConnectionError --rerun-except AssertionError`
* master: `1 failed, 2 rerun`
* this PR: `1 failed` (0 reruns)

`--only-rerun ConnectionError` alone still reruns on both branches, so it is specifically the combination — the one people write for "rerun network errors, never genuine assertion failures" — that breaks. Fixing the previous comment fixes this one too.

— _Comment created by Claude_

src/pytest_rerunfailures.py
```diff
@@ -545,12 +545,27 @@ def _matches_any_rerun_except_error(rerun_except_errors, excinfo):
     return _try_match_error(rerun_except_errors, excinfo)
 
 
+def _iter_exception_chain(exc):
+    seen = set()
+    while exc is not None and id(exc) not in seen:
+        seen.add(id(exc))
+        yield exc
+        if exc.__cause__ is not None:
```
icemac, line 611: **`ExceptionGroup` members are not part of the walked chain.**

The walk follows `__cause__`/`__context__` only, not `BaseExceptionGroup.exceptions`. Verified:

```python
def test_g():
    raise BaseExceptionGroup("boom", [MemoryError("out of memory")])
```

`pytest --reruns 2 --only-rerun MemoryError` → 0 reruns, on this branch as well as master.

Not a regression, but a gap in a change the README and changelog advertise as matching wrapped exceptions. This is the natural shape for `asyncio.TaskGroup` and `except*` code on 3.11+, and this plugin constructs a `BaseExceptionGroup` itself (line 994).

The rest of `_iter_exception_chain` looks right: `__cause__` takes priority, `__suppress_context__` is respected, and the `id()` set guards against cycles.

— _Comment created by Claude_

README.rst
```diff
@@ -110,6 +110,11 @@ would only rerun those errors that match ``AssertionError`` or ``ValueError``:
 
    $ pytest --reruns 5 --only-rerun AssertionError --only-rerun ValueError
 
+The same matching is applied to each exception in the ``__cause__`` /
+``__context__`` chain, so a wrapped error such as
+``raise RuntimeError(...) from MemoryError(...)`` is still rerun by
+``--only-rerun MemoryError``.
```
icemac, line 116: **Chain behavior is documented only under `--only-rerun`.**

The `--rerun-except` section immediately below is unchanged, even though that flag's semantics changed too — and in the surprising, rerun-suppressing direction described in my comment on `_try_match_error`. Whichever way that is resolved, this section should say so explicitly.

— _Comment created by Claude_

icemac (commented):
Automated review of this PR. Four findings; the first two are behavioral regressions I reproduced against `master`.

The root cause of both is that `_try_match_error` is shared by `_matches_any_only_rerun_error` and `_matches_any_rerun_except_error`, so walking the exception chain broadens *both* filters — and for `--rerun-except`, broader means fewer reruns.

— _Comment created by Claude_

src/pytest_rerunfailures.py
teddytennant: walks only explicit `__cause__` now, so an exception raised inside except/finally is not excluded by `--rerun-except` on the implicit context. pushed.

src/pytest_rerunfailures.py
teddytennant: same split; that combo reruns again.

src/pytest_rerunfailures.py
teddytennant: left ExceptionGroup alone, not a regression from this change.

README.rst
teddytennant: documented the `__cause__` vs `__context__` split under `--rerun-except`.

icemac (approved):
LGTM, thank you for the PR.

PR #365, opened 2026-09-08
Allow flaky conditions to inspect failure exceptions

### LINKED ISSUE #230: For "only_rerun" allow access exception attributes
I often have cases like below when I do not fail a test based on a specific exception attributes.

In case of `pytest-rerunfailures` the logic can look like:
```py
@pytest.mark.flaky(
    only_rerun=ClientResponseError,
    condition=lambda error: error.status in {429, 500},
)
async def test_request():
    await request("www.example.com")



@pytest.mark.flaky(
    only_rerun=CustomDatabaseError,
    condition=lambda error: error["code"] != 123456,
)
async def test_database():
    await database.set("foo", "bar")



def _callback(error: Exception) -> bool:
    if isinstance(error, CustomDatabaseError): ...
    if isinstance(error, AnotherError): ...
    raise 

@pytest.mark.flaky(
    only_rerun=[CustomDatabaseError, AnotherError],
    condition=_callback,
)
async def test_database():
    await database.set("foo", "bar")
```

There are also some 3rd-party packages that raise custom exceptions and I must handle an exception in different ways based on some attributes.

Question:  
Is this functionality out of scope of the current plugin and I should write a separate plugin?

ryux1:
Closes #230

This lets flaky conditions make one rerun decision for an entire test attempt from the exceptions that caused its failed phases:

- one-argument callable conditions receive each failure exception;
- existing zero-argument callable conditions remain supported;
- string conditions inspect the exception through the reserved `error` name;
- setup, call, teardown, and subtest failures retain their corresponding exceptions;
- an attempt is rerun when any failed phase matches, with each failure evaluated at most once.

`ExceptionInfo` stays on the worker-local test item, so xdist never serializes it, and the cached exception state is cleared after every attempt. Condition evaluation errors become non-fatal warnings even when warning filters promote warnings to errors.

Validation:

- pytest 8.2, 8.3, and 8.4: 206 passed, 19 skipped in each environment
- pytest 9.0 and 9.1: 216 passed, 9 skipped in each environment
- pytest 9.1 with pytest-xdist: 225 passed
- `pre-commit run --all-files` passed

### DIFF
--- a/docs/mark.rst
+++ b/docs/mark.rst
@@ -42,8 +42,10 @@ This will retry the test 5 times with a 2-second pause between attempts.
 ``condition``
 ^^^^^^^^^^^^^
 
-Re-run the test only if a specified condition is met.
-The condition can be any expression that evaluates to ``True`` or ``False``.
+Re-run the test only if a specified condition is met. The condition can be a
+boolean, a string to be evaluated, or a callable.
+
+Boolean conditions are evaluated directly:
 
 .. code-block:: python
 
@@ -55,6 +57,36 @@ The condition can be any expression that evaluates to ``True`` or ``False``.
        assert random.choice([True, False])
 
 In this example, the test will only be re-run if the operating system is Windows.
+
+A callable condition receives the exception that caused the test phase to fail.
+This allows a re-run decision to use exception attributes rather than only its
+type or message:
+
+.. code-block:: python
+
+   class TemporaryError(Exception):
+       def __init__(self, status):
+           self.status = status
+
+   @pytest.mark.flaky(
+       reruns=3,
+       condition=lambda error: error.status in {429, 503},
+   )
+   def test_service_request():
+       raise TemporaryError(429)
+
+A string condition can inspect the same exception through the ``error`` name.
+Its evaluation context also contains ``os``, ``sys``, ``platform``, ``config``
+(the pytest config object), and the test function's globals:
+
+.. code-block:: python
+
+   @pytest.mark.flaky(reruns=3, condition="error.status in {429, 503}")
+   def test_service_request():
+       raise TemporaryError(429)
+
+If a callable condition raises an exception, pytest emits a warning and does
+not re-run the test.
 
 
 ``only_rerun``
--- a/docs/mark.rst
+++ b/docs/mark.rst
@@ -58,7 +58,8 @@ Boolean conditions are evaluated directly:
 
 In this example, the test will only be re-run if the operating system is Windows.
 
-A callable condition receives the exception that caused the test phase to fail.
+A callable condition that accepts one argument receives the exception that
+caused a failed test phase. Existing zero-argument callables remain supported.
 This allows a re-run decision to use exception attributes rather than only its
 type or message:
 
@@ -75,9 +76,10 @@ type or message:
    def test_service_request():
        raise TemporaryError(429)
 
-A string condition can inspect the same exception through the ``error`` name.
-Its evaluation context also contains ``os``, ``sys``, ``platform``, ``config``
-(the pytest config object), and the test function's globals:
+A string condition can inspect the same exception through the reserved
+``error`` name. Its evaluation context also contains ``os``, ``sys``,
+``platform``, ``config`` (the pytest config object), and the test function's
+globals:
 
 .. code-block:: python
 
@@ -85,8 +87,10 @@ Its evaluation context also contains ``os``, ``sys``, ``platform``, ``config``
    def test_service_request():
        raise TemporaryError(429)
 
-If a callable condition raises an exception, pytest emits a warning and does
-not re-run the test.
+When more than one test phase fails in an attempt, the test is re-run if the
+condition matches any of those failures. Each failure is evaluated at most
+once. If a callable or string condition raises an exception, pytest emits a
+warning and does not re-run for that failure.
 
 
 ``only_rerun``
--- a/src/pytest_rerunfailures.py
+++ b/src/pytest_rerunfailures.py
@@ -1,5 +1,6 @@
 import hashlib
 import importlib.metadata
+import inspect
 import os
 import platform
 import re
@@ -302,22 +303,38 @@ def get_reruns_delay_backoff_factor(item):
     return factor
 
 
-def get_reruns_condition(item, excinfo=None, phase=None):
+def get_reruns_condition(item, failures=()):
     rerun_marker = _get_marker(item)
 
-    condition = True
-    if rerun_marker is not None and "condition" in rerun_marker.kwargs:
-        condition_results = getattr(item, "_rerun_condition_results", {})
-        if phase is not None and phase in condition_results:
-            return condition_results[phase]
-        condition = evaluate_condition(
-            item, rerun_marker, rerun_marker.kwargs["condition"], excinfo
-        )
-        if phase is not None:
-            condition_results[phase] = condition
-            item._rerun_condition_results = condition_results
+    if rerun_marker is None or "condition" not in rerun_marker.kwargs:
+        return True
 
-    return condition
+    condition = rerun_marker.kwargs["condition"]
+    condition_results = getattr(item, "_rerun_condition_results", {})
+    failures = list(failures)
+    if not failures:
+        failures = [("attempt", 0, None)]
+    if not callable(condition) and not isinstance(condition, str):
+        failures = failures[:1]
+
+    for phase, index, excinfo in failures:
+        cache_key = (phase, index)
+        if cache_key not in condition_results:
+            condition_results[cache_key] = evaluate_condition(
+                item, rerun_marker, condition, excinfo
+            )
+            item._rerun_condition_results = condition_results
+        if condition_results[cache_key]:
+            return True
+    return False
+
+
+def _warn_condition_error(msglines):
+    """Report a bad condition without letting warning filters abort pytest."""
+    try:
+        warnings.warn("\n".join(msglines))
+    except Warning:
+        pass
 
 
 def evaluate_condition(item, mark, condition: object, excinfo=None) -> bool:
@@ -328,13 +345,31 @@ def evaluate_condition(item, mark, condition: object, excinfo=None) -> bool:
     # Callable condition.
     if callable(condition):
         try:
-            return bool(condition(error))
+            try:
+                signature = inspect.signature(condition)
+            except (TypeError, ValueError):
+                call_with_error = True
+            else:
+                try:
+                    signature.bind(error)
+                except TypeError:
+                    try:
+                        signature.bind()
+                    except TypeError:
+                        _warn_condition_error([
+                            f"Error evaluating {mark.name!r} condition as a callable",
+                            "Condition callable must accept zero or one argument",
+                        ])
+                        return False
+                    call_with_error = False
+                else:
+                    call_with_error = True
+            return bool(condition(error) if call_with_error else condition())
         except Exception as exc:
-            msglines = [
+            _warn_condition_error([
                 f"Error evaluating {mark.name!r} condition as a callable",
                 *traceback.format_exception_only(type(exc), exc),
-            ]
-            warnings.warn("\n".join(msglines))
+            ])
             return False
 
     result = False
@@ -345,10 +380,10 @@ def evaluate_condition(item, mark, condition: object, excinfo=None) -> bool:
             "sys": sys,
             "platform": platform,
             "config": item.config,
-            "error": error,
         }
         if hasattr(item, "obj"):
             globals_.update(item.obj.__globals__)  # type: ignore[attr-defined]
+        globals_["error"] = error
         try:
             filename = f"<{mark.name} condition>"
             condition_code = compile(condition, filename, "eval")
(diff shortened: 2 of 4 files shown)


icemac:
The test failures against pytest-main are also on `master` and need to be fixed there before we can make progress here.

icemac:
`master` is fixed now, could you please resolve the conflicts?

src/pytest_rerunfailures.py
```diff
@@ -317,6 +337,7 @@ def evaluate_condition(item, mark, condition: object) -> bool:
             "sys": sys,
             "platform": platform,
             "config": item.config,
+            "error": error,
```
icemac: **A string condition that raises now aborts the entire run.**

(The actual code is a bit further down, in the `except Exception as exc` block of the string branch — around line 362 — which is outside the diff, so I'm anchoring here.)

That handler still ends in `fail(..., pytrace=False)`. Since `_should_not_rerun` is now reached from `pytest_runtest_protocol`, i.e. outside any test phase, the `Failed` exception escapes the hook:

```python
@pytest.mark.flaky(reruns=1, condition="error.status == 429")
def test_x():
    assert False
```

```
INTERNALERROR> Failed: Error evaluating 'flaky' condition
INTERNALERROR>     error.status == 429
INTERNALERROR> AttributeError: 'AssertionError' object has no attribute 'status'
no tests ran
```

This is not an edge case: introducing `error` makes eval-time `AttributeError` the *normal* outcome, because the attribute being probed only exists on some exception types. The callable branch handles the identical input gracefully (warn, return `False`); the string branch should do the same.

— _Comment created by Claude_

src/pytest_rerunfailures.py
```diff
@@ -294,21 +294,41 @@ def get_reruns_delay_backoff_factor(item):
...
+    # Callable condition.
+    if callable(condition):
+        try:
+            return bool(condition(error))
+        except Exception as exc:
+            msglines = [
+                f"Error evaluating {mark.name!r} condition as a callable",
+                *traceback.format_exception_only(type(exc), exc),
+            ]
+            warnings.warn("\n".join(msglines))
```
icemac: **This `warnings.warn` also crashes the run under `filterwarnings = error`.**

`_should_not_rerun` runs inside `pytest_runtest_protocol`, so the `UserWarning` is raised (not recorded) and propagates out of the hook. Your own `test_callable_condition_error_prevents_rerun` scenario plus a `pytest.ini` with `filterwarnings = error`:

```
INTERNALERROR> UserWarning: Error evaluating 'flaky' condition as a callable
INTERNALERROR> ValueError: nope
```

`filterwarnings = error` is a very common setting. Consider reporting via `item.warn(...)` or catching the raised warning here.

— _Comment created by Claude_

src/pytest_rerunfailures.py
```diff
@@ -1043,6 +1077,13 @@ def pytest_runtest_makereport(item, call):
         # create a dict to store xfail results for each stage
         setattr(item, "_test_xfailed", {})
 
+        # Keep exception state on the worker-side item. TestReport attributes
+        # are serialized by pytest-xdist and ExceptionInfo is not serializable.
+        setattr(item, "_rerun_condition_excinfo", {})
+        setattr(item, "_rerun_condition_results", {})
+
+    item._rerun_condition_excinfo[result.when] = call.excinfo
```
icemac: **This unconditional assignment clobbers subtest exceptions, so the feature never fires for failed subtests.**

With `pytest-subtests`, `pytest_runtest_makereport` fires twice for `when == "call"`: first for the subtest report (real `excinfo`), then for the aggregate report where `call.excinfo is None`. Verified with a trylast wrapper:

```
MAKEREPORT when=call failed=True  excinfo=<ExceptionInfo ValueError('boom') tblen=1>
MAKEREPORT when=call failed=False excinfo=None
```

So `condition=lambda error: isinstance(error, ServiceError) and error.status == 429` on a test whose subtest raises `ServiceError(429)` sees `error=None` and is **not** re-run — even though re-running on failed subtests is explicitly supported (`has_failed_subtests`, #357).

The `_test_xfailed` line right below already solves this by accumulating rather than replacing; the same shape works here:

```python
if call.excinfo is not None or result.when not in item._rerun_condition_excinfo:
    item._rerun_condition_excinfo[result.when] = call.excinfo
```

— _Comment created by Claude_

src/pytest_rerunfailures.py
```diff
@@ -565,18 +586,20 @@ def _should_hard_fail_on_error(item, report, excinfo):
...
         or (not report.failed and not has_failed_subtests)
         or xfail
         or is_terminal_error
-        or not condition
-    )
+    ):
+        return True
+
+    excinfo = item._rerun_condition_excinfo.get(report.when)
+    return not get_reruns_condition(item, excinfo, report.when)
```
icemac: **Evaluating the condition per phase lets the phases disagree, and a test can end up reported as both FAILED and PASSED.**

The condition result is now keyed by `report.when`, so the call phase can answer "don't re-run" while the teardown phase answers "re-run". The call report is then logged as a final failure, and the teardown report still triggers the re-run. With a call phase failing on `AssertionError`, a function-scoped fixture whose teardown raises `ServiceError(503)`, and `condition=lambda e: getattr(e, "status", None) == 503`:

```
test_a.py::test_x FAILED
test_a.py::test_x RERUN
test_a.py::test_x PASSED
1 failed, 2 passed, 1 rerun
```

Before this PR the condition was a single item-wide value, so the two reports could not disagree. Related: because `pytest_runtest_teardown` evaluated the *call*-phase condition as falsy, the higher-scope finalizers were not held back for a re-run that then happened anyway — the exact situation #357 was fixed to avoid.

Worth deciding explicitly what the intended semantics are: one decision per attempt (e.g. "re-run if any failed phase's condition matches"), rather than an independent decision per report.

— _Comment created by Claude_

src/pytest_rerunfailures.py
```diff
@@ -317,6 +337,7 @@ def evaluate_condition(item, mark, condition: object) -> bool:
             "sys": sys,
             "platform": platform,
             "config": item.config,
+            "error": error,
         }
         if hasattr(item, "obj"):
             globals_.update(item.obj.__globals__)  # type: ignore[attr-defined]
```
icemac, line 385: **This `update()` silently shadows the new `error` name.**

`globals_["error"]` is set three lines up and then overwritten by anything named `error` in the test module (a variable, or `from x import error`):

```python
error = None  # module level

@pytest.mark.flaky(reruns=1, condition="isinstance(error, ServiceError)")
def test_x():
    raise ServiceError(429)
```

→ `1 failed`, no re-run; the condition saw the module global. Setting `"error"` *after* the `update()` makes the documented name unshadowable.

— _Comment created by Claude_

src/pytest_rerunfailures.py
```diff
@@ -294,21 +294,41 @@ def get_reruns_delay_backoff_factor(item):
...
 
 
-def evaluate_condition(item, mark, condition: object) -> bool:
+def evaluate_condition(item, mark, condition: object, excinfo=None) -> bool:
     # copy from python3.8 _pytest.skipping.py
 
+    error = excinfo.value if excinfo is not None else None
+
+    # Callable condition.
+    if callable(condition):
```
icemac, line 346: **This is a breaking change for existing zero-argument callable conditions, and it fails silently.**

Checking `callable(condition)` before the boolean branch changes behaviour for anyone who passed a function today: it used to fall through to `bool(condition)` — always truthy, i.e. "always re-run". Now it is invoked with one positional argument.

```python
@pytest.mark.flaky(reruns=1, condition=lambda: sys.platform == "darwin")
```

- master: `1 passed, 1 rerun`
- this branch: `1 failed, 1 warning` — the `TypeError` is swallowed into a generic warning and re-runs are silently off.

Two suggestions:
1. The changelog fragment is `changes/230.feature.rst`; towncrier is configured with a `breaking` category, which this warrants.
2. An arity mismatch deserves a specific message ("condition callable must accept one argument") rather than the same generic warning used for exceptions raised *inside* the condition — otherwise the failure mode is hard to diagnose.

— _Comment created by Claude_

src/pytest_rerunfailures.py
```diff
@@ -927,6 +950,16 @@ def _is_rerun_path_excluded(item):
     )
 
 
+def _get_reruns_condition_failure(item):
+    """Return the phase and exception for the most recent failed test phase."""
+    failed_statuses = getattr(item, "_test_failed_statuses", {})
+    excinfos = getattr(item, "_rerun_condition_excinfo", {})
+    for phase in ("teardown", "call", "setup"):
```
icemac: **The `"teardown"` iteration is unreachable, and the `(None, None)` return defeats the caching this PR adds.**

This helper is only called from `pytest_runtest_teardown`, which runs *before* the teardown report is created, so `_test_failed_statuses["teardown"]` is never set at this point.

When only subtests failed (the call phase itself passing), it returns `(None, None)`, and `get_reruns_condition` with `phase=None` doesn't consult or populate the cache — so the condition runs twice in a single attempt. With a counting callable condition on a failing subtest, `COND ... None` is printed twice per attempt, which contradicts the "each phase condition is evaluated once" claim in the PR description.

Minor: the docstring says "most recent failed test phase", but the order here is fixed, not chronological.

— _Comment created by Claude_

src/pytest_rerunfailures.py
```diff
@@ -1043,6 +1077,13 @@ def pytest_runtest_makereport(item, call):
         # create a dict to store xfail results for each stage
         setattr(item, "_test_xfailed", {})
 
+        # Keep exception state on the worker-side item. TestReport attributes
+        # are serialized by pytest-xdist and ExceptionInfo is not serializable.
+        setattr(item, "_rerun_condition_excinfo", {})
```
icemac: **These dicts keep an `ExceptionInfo` alive on every item for the whole session.**

`_rerun_condition_excinfo` is populated for *all* items, flaky or not, and is only reset at the next `setup` of the same item — which never happens for the final attempt. Since `session.items` holds every item until the run ends, each failed test's full traceback (and therefore all of its frame locals) stays reachable. pytest deliberately keeps only the string `longrepr` on reports for exactly this reason.

Clearing the dict once the item's re-run decision is final (or only recording it when a flaky marker with a `condition` is present) would avoid the retention.

— _Comment created by Claude_

docs/mark.rst
```diff
@@ -56,6 +58,36 @@ The condition can be any expression that evaluates to ``True`` or ``False``.
...
+Its evaluation context also contains ``os``, ``sys``, ``platform``, ``config``
+(the pytest config object), and the test function's globals:
+
+.. code-block:: python
+
+   @pytest.mark.flaky(reruns=3, condition="error.status in {429, 503}")
+   def test_service_request():
+       raise TemporaryError(429)
+
+If a callable condition raises an exception, pytest emits a warning and does
```
icemac: This sentence doesn't hold in two cases, both of which produce `INTERNALERROR` rather than a warning: string conditions that raise during evaluation, and callable conditions under `filterwarnings = error` (see the two inline comments in `src/pytest_rerunfailures.py`). Once those are fixed the docs will be accurate as written.

Separately, the `error` name is documented for string conditions without mentioning that a module-level `error` in the test file overrides it.

— _Comment created by Claude_

icemac (commented):
Thanks for working on #230 — the xdist-safe approach of keeping `ExceptionInfo` on the worker-side item is the right idea. A few things need fixing before this can go in; two of them abort the whole run with `INTERNALERROR`, and one makes the feature a no-op for failed subtests. Details inline.

— _Comment created by Claude_

src/pytest_rerunfailures.py
ryux1: 04cb3ef makes string evaluation failures non-fatal warnings and returns false for that failure. The regression covers a missing exception attribute and asserts that pytest does not exit with INTERNAL_ERROR.

src/pytest_rerunfailures.py
ryux1: 04cb3ef routes condition warnings through a helper that catches warning exceptions, so `filterwarnings = error` cannot escape the protocol hook. The new regression exercises that exact configuration.

src/pytest_rerunfailures.py
ryux1: 04cb3ef accumulates only failing, non-null ExceptionInfo values per phase instead of overwriting them. The new pytest-subtests regression verifies the callable receives ServiceError(429) exactly once.

src/pytest_rerunfailures.py
ryux1: 04cb3ef now computes one attempt-level decision after all reports: any matching failed phase reruns the attempt. Nonmatching earlier failures are not published as final when a later phase carries the rerun, and finalizers remain suspended until the complete decision is known. The mixed call/teardown regression exits cleanly with one pass and one rerun.

src/pytest_rerunfailures.py
ryux1: 04cb3ef applies test globals first and then assigns the reserved `error` binding. The regression defines a module-level `error = None` and verifies the actual ServiceError still drives the rerun.

src/pytest_rerunfailures.py
ryux1: 04cb3ef replaces the phase-selection helper with an accumulated failed-exception list and one cached attempt decision. The subtest regression verifies its exception is evaluated once, and teardown failures are included once their report exists.

src/pytest_rerunfailures.py
ryux1: 04cb3ef clears both the ExceptionInfo collection and its result cache at the end of every attempt. A regression inspects the collected item from the following test and verifies both stores are empty.

docs/mark.rst
ryux1: 04cb3ef updates the docs to cover callable and string failures, the reserved unshadowable `error` binding, zero-argument compatibility, and the any-matching-failure attempt semantics. Both INTERNAL_ERROR cases now have regressions.

src/pytest_rerunfailures.py
ryux1: 04cb3ef preserves zero-argument callable conditions by inspecting the callable signature; one-argument callables receive the exception. Unsupported arities get a specific zero-or-one-argument warning. The changelog remains a feature because the existing zero-argument behavior is retained.

icemac (approved):

PR #367, opened 2026-09-10
Preserve rerun status for teardown failures

### LINKED ISSUE #366: Teardown error persists after successful rerun when test and teardown both fail initially
### Description

With the current pytest-rerunfailures v16.6.1, if both the test and teardown fail during the initial execution, the test is reported as failing even though the rerun passes. The issue is reproducible with the 16.6.1 but not the 16.6 version.
This also does not happen if only one of the two (either teardown or test) fails.

### Example
```python
import pytest

@pytest.fixture
def demo_fixture(request):
    yield
    if request.node.execution_count == 1:
        raise RuntimeError("Some failure")

def test_demo(demo_fixture, request):
    if request.node.execution_count == 1:
        assert False, "Test failure"
```

Results in 
```
uv run pytest .\test_teardown_fail.py --reruns 1
===== test session starts =====
platform win32 -- Python 3.14.7, pytest-9.1.1, pluggy-1.6.0
rootdir: C:\repos\pytest-rerun-bug
configfile: pyproject.toml
plugins: rerunfailures-16.6.1
collected 1 item                                                                                                                                                                                                                                                                              

test_teardown_fail.py RE.                                                                                                                                                                                                                                                               [100%]
[...]
===== 1 passed, 1 error, 1 rerun in 0.14s =====
```
and an exit code of `1`

LouisDeconinck:
Fixes #366

When a test phase triggers a rerun, a later failed phase from the same attempt must also be reported as a rerun. Previously, a teardown failure after a failed test remained an error even when the rerun passed.

This changes only failed later-phase reports; successful reports keep their existing outcome. Added the issue reproducer, updated the persistent-teardown expectation, and added the required bugfix changelog fragment.

Checks:
- `uv run pytest -q` (217 passed, 9 skipped)
- `uv run --group dev mypy --ignore-missing-imports src/pytest_rerunfailures.py`
- `uv run --with pre-commit pre-commit run --all-files`
- `git diff --check`


icemac (approved):

icemac:
Thank you for this PR. 😃

PR #369, opened 2026-09-11
Add pytest.ini support for only_rerun

### LINKED ISSUE #165: Add pytest.ini support for --reruns and --only-rerun options
Currently the options are only available on the command line (and thus must be provided every time).   If support were added to read these settings from a pytest.ini file, the developer wouldn't need to remember to add them every time.  It can be done indirectly through the "addopts" arguments, but it would be more clear/self-documenting if they were distinct options as well.

LouisDeconinck:
Fixes #165.

Add the missing pytest.ini configuration for --only-rerun while preserving command-line precedence. The setting accepts one regex per line, matching the existing repeatable CLI option.

Included a regression test, configuration documentation, and a towncrier feature entry.


tests/test_pytest_rerunfailures.py
```diff
@@ -1107,6 +1107,28 @@ def test_only_rerun2():
     )
 
 
+def test_only_rerun_ini(testdir):
```
janmrow, line 1110: Could we also cover multiple `only_rerun` entries here? 

The docs support one expression per line, but this test only covers a single entry. For example, configuring both AssertionError and ValueError and expecting both to be rerun would cover the multi-line INI behavior.

LouisDeconinck:
Added coverage for multiple `only_rerun` entries.

tests/test_pytest_rerunfailures.py
LouisDeconinck: Done — added `test_only_rerun_ini_multiple`, which configures a multi-line `only_rerun` INI value with both `AssertionError` and `ValueError`. Both tests are rerun while a `KeyError` test still fails without a rerun. Verified: 3 failed, 2 reruns.

icemac (commented):
Thanks for picking this up! Precedence (CLI beats ini), `pyproject.toml` support, `-o only_rerun=` and `--help` rendering all check out. A few remarks inline, mostly about the sibling option, tests and docs.

— _Comment created by Claude_

src/pytest_rerunfailures.py
```diff
@@ -189,6 +189,11 @@ def pytest_addoption(parser):
         RERUNS_DELAY_BACKOFF_FACTOR_DESC,
         type=arg_type,
     )
+    parser.addini(
```
icemac, line 198: Only `only_rerun` gets an ini option; its documented sibling `rerun_except` stays CLI-only. With `rerun_except = ValueError` in `pytest.ini` the setting is silently ignored, the only signal being `PytestConfigWarning: Unknown config option: rerun_except` in the warnings summary. Since the docs introduce the two as a mirrored pair, adding `parser.addini("rerun_except", ..., type="linelist")` here would keep them in sync and also lets the fallback below drop its hardcoded name check.

— _Comment created by Claude_

src/pytest_rerunfailures.py
```diff
@@ -189,6 +189,11 @@ def pytest_addoption(parser):
...
+        "only_rerun",
+        "only rerun errors matching the regex provided.",
```
icemac: The other three `addini` calls share their help text with the CLI option via the `*_DESC` constants (`RERUNS_DESC` etc.) so the two cannot drift. This one inlines a paraphrase that drops the accumulation sentence, so `pytest --help` gives no hint that the ini form takes one regex per line (which `docs/configuration.rst` does explain). Suggest hoisting an `ONLY_RERUN_DESC` next to the other constants and using it for both `--only-rerun` and this `addini`.

— _Comment created by Claude_

src/pytest_rerunfailures.py
```diff
@@ -189,6 +189,11 @@ def pytest_addoption(parser):
         RERUNS_DELAY_BACKOFF_FACTOR_DESC,
         type=arg_type,
     )
+    parser.addini(
+        "only_rerun",
```
icemac, line 199: Not new to this PR, but worth considering now that the pattern can live in a checked-in file: an invalid regex such as `only_rerun = [unclosed` produces `INTERNALERROR> re.error: unterminated character set at position 0` with a traceback into `re/_parser.py` and nothing naming `only_rerun` or the config file. A CLI typo breaks the one command you typed; a committed ini typo breaks every run in CI. `check_options` already raises `pytest.UsageError` for other option problems and would be the natural place to `re.compile` each entry.

— _Comment created by Claude_

src/pytest_rerunfailures.py
```diff
@@ -591,6 +596,8 @@ def _get_rerun_filter_regex(item, regex_name):
             regex = [regex]
     else:
         regex = getattr(item.session.config.option, regex_name)
+        if regex is None and regex_name == "only_rerun":
+            regex = item.session.config.getini("only_rerun")
```
icemac: `_get_rerun_filter_regex` is generic over `regex_name` (it is called with both `"only_rerun"` and `"rerun_except"`) and reads the CLI value generically via `getattr`, but this fallback hardcodes the name and re-types the literal in `getini()`. It also makes the return type differ per name (`[]` for `only_rerun`, `None` for `rerun_except`). If both names are registered with `addini`, this can simply be:

```python
if regex is None:
    regex = item.session.config.getini(regex_name)
```

— _Comment created by Claude_

tests/test_pytest_rerunfailures.py
```diff
@@ -1107,6 +1107,55 @@ def test_only_rerun2():
...
```
icemac, line 1110: The central claim of the PR, that the CLI still overrides the ini, has no test. The file pairs `test_ini_file_parameters` with `test_ini_file_parameters_override` for the existing ini options; this new option only gets the plain half. I confirmed manually that ini `only_rerun = AssertionError` plus `--only-rerun ValueError` reruns only the ValueError test, but a refactor that swaps the lookup order (or uses `getattr(...) or getini(...)`) would invert precedence and the suite would stay green. Also worth a test: marker `only_rerun` vs ini, and ini `only_rerun` combined with `--rerun-except`, which takes the four-way branch in `_should_hard_fail_on_error`.

— _Comment created by Claude_

docs/configuration.rst
```diff
@@ -35,6 +35,21 @@ Below are the ``pytest.ini`` options supported by the plugin:
      [pytest]
      reruns_delay = 2.5
 
+``only_rerun``
+^^^^^^^^^^^^^^
+
+- **Description**: Sets regular expressions for errors that should be rerun. Add one expression per line.
```
icemac: Two behaviours of the new ini tier are not stated anywhere: (a) `--only-rerun` on the command line *replaces* the ini list rather than accumulating with it, even though the CLI help says "pass this flag multiple times to accumulate"; (b) a repo-wide `only_rerun` also applies to tests marked plain `@pytest.mark.flaky(reruns=2)`, so a test raising a non-matching error stops rerunning with no warning. Both match existing `--only-rerun` semantics, but a checked-in ini key becomes the permanent project default. The Priority sections in `docs/mark.rst` and `README.rst` still describe precedence as marker-vs-CLI only and should mention the ini tier.

— _Comment created by Claude_

docs/configuration.rst
```diff
@@ -35,6 +35,21 @@ Below are the ``pytest.ini`` options supported by the plugin:
      [pytest]
      reruns_delay = 2.5
 
+``only_rerun``
```
icemac, line 38: `README.rst` is also the PyPI long description (see `readme` in `pyproject.toml`), and it is not touched here: the "Re-run all failures matching certain expressions" section still presents `--only-rerun` as CLI-only and the "Priority" section shows the `pytest.ini` layer holding only `reruns = 3`. Leaving it stale reproduces the confusion that led to #165. Previous option additions (e.g. `--reruns-delay-backoff-factor`) updated README together with the docs.

— _Comment created by Claude_

docs/configuration.rst
```diff
@@ -45,6 +60,7 @@ To configure your test environment for consistent retries and delays, add the fo
    [pytest]
    reruns = 3
    reruns_delay = 2.0
+   only_rerun = AssertionError
```
icemac, line 78: The example gained a third key, but the lead-in above ("for consistent retries and delays") and the "This setup ensures that:" list below still cover only two. A third bullet would help, especially since `only_rerun` *restricts* reruns rather than adding behaviour.

— _Comment created by Claude_

LouisDeconinck:
All remarks addressed in 9e9a542:

- `rerun_except` is now a registered ini option (`type="linelist"`), mirroring `only_rerun` — the fallback in `_get_rerun_filter_regex` is now a plain `getini(regex_name)` for both names.
- Hoisted `ONLY_RERUN_DESC` / `RERUN_EXCEPT_DESC` shared by the CLI flags and `addini`, mentioning the one-per-line ini form.
- `check_options` now compiles each CLI/ini pattern and raises `UsageError` naming the option, so a committed `[unclosed` fails fast with a clear message instead of an INTERNALERROR traceback.
- Tests: CLI-overrides-ini precedence (asserts the R-marker lands on the ValueError test), marker-overrides-ini, ini `only_rerun` combined with `--rerun-except` exercising the four-way branch, `rerun_except` from ini, and invalid-regex cases for both names.
- Docs: README gained the ini forms in both filter sections plus the marker > CLI > ini order in Priority; configuration.rst documents `rerun_except`, the replace-not-accumulate behavior, and the marker-interaction caveat; mark.rst mentions the ini tier.

234 tests pass; ruff and sphinx-lint clean.

icemac (approved):

PR #370, opened 2026-09-11
Show each rerun attempt outcome in summary

### LINKED ISSUE #191: Enhancement: present final `rerun test summary info` section in the same way as the `test session starts` section
The `test session starts` section is really nice in how it presents a running test suite's test outcomes: each test is shown in succession, is color-coded according to outcome, and the whole flow is easy to follow because you can follow along as individual tests get run, and either pass or fail, thereby triggering the plugin's rerun mechanism.

For example, it's easy to see in the following run:
- that the first test passed without a rerun
- the second test ran twice (1 rerun) but failed both times, leading to an overall failed test result
- the third test ran three times (2 reruns) and eventually passed
- the fourth test ran four times (3 reruns) and eventually failed
- the fifth test ran three times (2 reruns) and eventually failed
- the last test ran once and passed

Screen Shot 2022-05-17 at 11 49 05 PM

However, the `rerun_test_summary_info` section (which is obtained by invoking command line option `-raR`) leaves a bit to be desired:
- not color-coded
- only states which tests required a rerun (regardless of final outcome) instead of giving a blow-by-blow account of each test as reruns occur

Is it possible to make the final section more like the first? 

Here is my testcase for above screenshot. I ran the test session as:
`pytest -v -raR -k flaky`

```
import pytest
import random

def flakey_pass_fail():
    random.seed(random.randint(0,100))
    return random.randint(0, 1) == 1

@pytest.mark.flaky(reruns=0)
def test_flaky_0_reruns():
    result = bool(flakey_pass_fail())
    assert random.randint(0,1)

@pytest.mark.flaky(reruns=1)
def test_flaky_1_reruns():
    result = bool(flakey_pass_fail())
    assert random.randint(0,1)

@pytest.mark.flaky(reruns=2)
def test_flaky_2_reruns():
    result = bool(flakey_pass_fail())
    assert random.randint(0,1)

@pytest.mark.flaky(reruns=3)
def test_flaky_3_reruns():
    result = bool(flakey_pass_fail())
    assert random.randint(0,1)

@pytest.mark.flaky(reruns=2)
def test_flaky_2_reruns_but_always_fail():
    assert False

@pytest.mark.flaky(reruns=2)
def test_flaky_2_reruns_and_always_pass():
    assert True
```

LouisDeconinck:
Fixes #191. The rerun summary now preserves one report per test attempt and displays each rerun and final outcome with pytest-style status colors. Tests: .venv/bin/pytest -q; pre-commit run --all-files --show-diff-on-failure.


src/pytest_rerunfailures.py
```diff
@@ -705,6 +705,15 @@ def is_master(config):
     return not (hasattr(config, "workerinput") or hasattr(config, "slaveinput"))
 
 
+class RerunSummary:
+    def __init__(self):
+        self.reports = []
+
+    def pytest_runtest_logreport(self, report):
+        if hasattr(report, "rerun"):
```
icemac: **xdist crash reruns vanish from the summary.**

xdist's `handle_crashitem` synthesises the crash report with `when="???"` and without a `rerun` attribute; `XDistHooks.pytest_handlecrashitem` only flips `outcome`. So this `hasattr` filter drops the report, the nodeid never lands in `rerun_nodeids`, and the whole test is suppressed in `show_rerun`. (Even with `rerun` set, `when="???"` is rejected by the `when not in (...)` check below.)

Reproduced with `pytest -n 2 --reruns 2 -r R` on a test calling `os._exit(1)`: master prints two `RERUN` lines, this branch prints no rerun section at all. With a second, normally-flaky test present the section appears, the footer still says `2 rerun`, but the crashed test is silently missing. `test_rerun_passes_after_temporary_test_crash` runs with `-r R` but only asserts outcomes, so CI does not catch this.

— _Comment created by Claude_

icemac (commented):
Review of the rerun-summary change. Findings 1–3 (xdist crash reruns, teardown tracebacks, xdist ordering) were reproduced by running the plugin on master vs. this branch; the full suite passes on the branch because the existing summary tests only assert outcomes.

— _Comment created by Claude_

src/pytest_rerunfailures.py
```diff
@@ -1337,20 +1347,74 @@ def pytest_terminal_summary(terminalreporter):
...
+    )
+    if summary is not None:
+        attempts = {}
+        rerun_nodeids = set()
+        for report in summary.reports:
+            if report.outcome == "rerun":
+                rerun_nodeids.add(report.nodeid)
+            if report.when not in ("setup", "call", "teardown"):
+                continue
+            key = (report.nodeid, report.rerun)
```
icemac: **Keeping one report per `(nodeid, rerun)` drops the teardown traceback.**

When call and teardown both fail in the same attempt, both reports get `outcome="rerun"`, but `_rerun_summary_priority` prefers `when == "call"` and tracebacks are only emitted for the winner.

Reproduced with a failing test body plus a fixture whose teardown raises, run with `--reruns 1 --rerun-show-tracebacks`: master shows the teardown traceback twice, this branch shows it zero times. That contradicts the README ("Each retried attempt's traceback is appended...") and regresses #367. Side effects: the `RERUN` line count no longer matches the `N rerun` footer, and the final attempt's teardown `ERROR` is hidden behind the call's `FAILED` line.

— _Comment created by Claude_

src/pytest_rerunfailures.py
```diff
@@ -1337,20 +1347,74 @@ def pytest_terminal_summary(terminalreporter):
...
+            current = attempts.get(key)
+            if current is None or _rerun_summary_priority(
+                report
+            ) > _rerun_summary_priority(current):
+                attempts[key] = report
+
+        for report in attempts.values():
```
icemac: **Output order depends on report arrival order, which xdist shreds.**

Reproduced with two flaky tests under `-n 2 --reruns 3 -r R`:

```
RERUN test_b
RERUN test_a
RERUN test_b
RERUN test_a
PASSED test_a
PASSED test_b
```

The old flat `RERUN`-only list did not care about order; the new grouped `RERUN/RERUN/PASSED` format only makes sense when each test's attempts are contiguous. Suggest emitting grouped by nodeid (first-seen order), then by `report.rerun`.

— _Comment created by Claude_

src/pytest_rerunfailures.py
```diff
@@ -1337,20 +1347,74 @@ def pytest_terminal_summary(terminalreporter):
 
     lines = show_rerun(terminalreporter, show_tracebacks=show_tracebacks)
     if lines:
-        tr._tw.sep("=", "rerun test summary info")
+        tr.write_sep("=", "rerun test summary info", cyan=True, bold=True)
         for line in lines:
-            tr._tw.line(line)
+            status, _, _ = line.partition(" ")
```
icemac: **Traceback lines are re-parsed as status words and colourised.**

`show_rerun` flattens `str(report.longrepr).splitlines()` into the same list as the status lines, so this lookup runs on every traceback line too. With `--tb=native` an `AssertionError("boom\nERROR while doing thing")` renders the second line red; a fixture named `ERROR` gets its `ERROR = 'x'` line coloured as well.

Cosmetic, but the root cause is the format-then-reparse round trip. `show_rerun` could return `(status_or_None, text)` pairs instead.

— _Comment created by Claude_

src/pytest_rerunfailures.py
```diff
@@ -1337,20 +1347,74 @@ def pytest_terminal_summary(terminalreporter):
...
+            if show_tracebacks and rep.longrepr:
+                lines.extend(str(rep.longrepr).splitlines())
     return lines
 
 
+def _rerun_summary_priority(report):
+    return (report.outcome in ("failed", "rerun"), report.when == "call")
+
+
+def _rerun_summary_status(report):
```
icemac: **This re-implements `pytest_report_teststatus`.**

pytest's own hookimpls already return `ERROR` for failed setup/teardown, `SKIPPED`, `XFAIL`/`XPASS`, and `_pytest/terminal.py` holds the same passed/failed/skipped markup table. This plugin's own `pytest_report_teststatus` already returns `("RERUN", {"yellow": True})`, so the dict above is a second hand-maintained copy of that markup.

A single `tr.config.hook.pytest_report_teststatus(report=report, config=tr.config)` call (unpacking the optional `(word, markup)` tuple as `_pytest/terminal.py` does) replaces both this function and the dict literal, and also honours third-party outcome categories. The dict literal is currently rebuilt on every line as well.

— _Comment created by Claude_

src/pytest_rerunfailures.py
```diff
@@ -705,6 +705,15 @@ def is_master(config):
     return not (hasattr(config, "workerinput") or hasattr(config, "slaveinput"))
 
 
+class RerunSummary:
```
icemac: **`RerunSummary` duplicates data `terminalreporter.stats` already holds.**

`TerminalReporter.pytest_runtest_logreport` calls `_add_stats` before its early return, so every report, including passed setup/teardown under the `""` category, is already in `terminalreporter.stats`. `show_rerun` can get the same set via `chain.from_iterable(terminalreporter.stats.values())` filtered on `hasattr(rep, "rerun")`, and sorting that by `(nodeid, rerun)` fixes the xdist ordering issue too.

As written, this plugin is registered on every run (including every xdist worker, which never emits a summary) and, since `report.rerun` is set on all three phase reports, keeps three references per test for the whole session.

— _Comment created by Claude_

src/pytest_rerunfailures.py
```diff
@@ -1337,20 +1347,74 @@ def pytest_terminal_summary(terminalreporter):
...
+                attempts[key] = report
+
+        for report in attempts.values():
+            if report.nodeid not in rerun_nodeids:
+                continue
+            status = _rerun_summary_status(report)
+            lines.append(f"{status} {report.nodeid}")
+            if show_tracebacks and status == "RERUN" and report.longrepr:
+                lines.extend(str(report.longrepr).splitlines())
+    else:
```
icemac: **This fallback is unreachable and already diverges from the main path.**

`pytest_configure` always registers the summary plugin under a fixed name, so `get_plugin` cannot return `None` in practice. The branch is not covered by the suite and already behaves differently (it emits tracebacks unconditionally; the new path only for `RERUN` lines). Note that `-p no:pytest-rerunfailures-summary` silently reverts to the old format with no warning. Either drop the branch (it disappears if `terminalreporter.stats` is used directly) or add a test for it.

— _Comment created by Claude_

tests/test_pytest_rerunfailures.py
```diff
@@ -537,6 +538,33 @@ def test_pass():
...
+        def test_eventually_passes():
+            global attempts
+            attempts += 1
+            assert attempts == 3
+        """
+    )
+    result = testdir.runpytest("--reruns", "2", "-r", "R", "--color", "yes")
+
+    stdout = result.stdout.str()
+    assert "\x1b[33mRERUN " in stdout
```
icemac: **Only the `RERUN`/`PASSED` path is covered, and the colour assertions match whole stdout.**

`ERROR`, `SKIPPED`, `XFAIL`, `XPASS`, the call-vs-teardown tie-break in `_rerun_summary_priority`, and the `else` fallback have no coverage. `ERROR`/`SKIPPED` are reachable today and cheap to add.

These two `\x1b[...` assertions are also false-negative-prone: with `PYTEST_ADDOPTS=-v` the verbose progress line already contains a green `PASSED`, so they pass even if the rerun section emits no colour. The de-ANSI'd `statuses == [...]` check below already proves the statuses; scope the colour check to the section.

— _Comment created by Claude_

LouisDeconinck:
Thanks for the thorough review — all findings reproduced and addressed in 97fb513:

- **xdist crash reruns vanish**: `pytest_handlecrashitem` now stamps `report.rerun` with the attempt index (from `get_test_failures`, before `add_test_failure`), so crash reports are collected and ordered alongside the rescheduled attempt's own reports. `pytest -n 2 --reruns 2 -r R` on an `os._exit(1)` test now prints `RERUN`, `RERUN`, `FAILED` — matching the `2 rerun` footer.
- **Dropped teardown traceback / hidden final ERROR**: the `(nodeid, rerun)` dedup is gone. Each emitted report gets its own line (sorted by `rerun`, then setup/call/teardown), so a call+teardown failure in one attempt shows both `RERUN` lines with both tracebacks under `--rerun-show-tracebacks`, and the final attempt shows `FAILED` + `ERROR`.
- **xdist ordering**: reports are grouped by nodeid (first-seen order) then sorted by attempt number, so interleaved worker reports render as contiguous per-test groups.
- **Traceback lines re-colourised**: `show_rerun` now returns `(line, markup)` pairs; traceback lines carry `None` markup and are written plainly.
- **Re-implemented `pytest_report_teststatus`**: status words and markup now come from `config.hook.pytest_report_teststatus` (with the same default outcome colouring `_pytest/terminal.py` applies for plain-string words), so `ERROR`, `SKIPPED`, `XFAIL`/`XPASS`, and third-party categories are all handled consistently.
- **`RerunSummary` duplicates `terminalreporter.stats`**: the plugin is removed entirely; `show_rerun` iterates `chain.from_iterable(terminalreporter.stats.values())` filtered on `hasattr(report, "rerun")`, which also fixes the ordering issue per your suggestion.
- **Unreachable `get_plugin` fallback**: gone with the plugin removal — `-p no:pytest-rerunfailures-summary` no longer exists as a silent revert path.
- **Coverage**: added tests for the `ERROR` (failed setup) and `SKIPPED` (skipped call + failing teardown) paths, call+teardown traceback preservation, the crash summary line, and xdist per-test grouping; the colour assertions are now scoped to the rerun section.

On XFAIL/XPASS specifically: xfail-marked tests never reach the summary in practice — `_should_not_rerun` excludes `wasxfail` reports, and pytest converts teardown failures on xfail-marked tests into xfail too. Delegating to the hook still honours those (and any plugin-specific) statuses if they ever appear.

Verified locally: 248 tests pass on pytest 9.1.1 + xdist 3.8.0, ruff check/format clean.

icemac (approved):
LGTM

PR #371, opened 2026-09-11
Clarify scoped fixture rerun support

### LINKED ISSUE #183: Clarification of class/module/package level fixture support
It is prominently listed on the main page that:

> This plugin may not be used with class, module, and package level fixtures.

However as part of #66 certain issues with these fixtures were resolved as well as some tests added. So it seems to me that there is at least *some* level of support for these types of fixtures albeit undocumented what is the expected behavior.
I'm currently looking into this library as a replacement for https://github.com/box/flaky, which completely lacks the option to rerun tests on failed fixtures.

So maybe some of these questions could be answered as part of the docs:

- What is the current status of fixtures in this project?
- Are there plans in adding further support?
- Would pull-requests for these types of fixtures be accepted?

LouisDeconinck:
Fixes #183.

Clarify the currently supported fixture behavior in the README: setup failures and `setup_class` failures are rerun for class-, module-, package-, and session-scoped fixtures, while successful higher-scoped fixtures remain cached across a rerun and are torn down once afterward.

Add a package-scoped setup regression test and a towncrier documentation fragment.

Checks:
- `PYTHONPATH=src python3 -m pytest -q` (227 passed, 9 skipped)
- `uvx --from pre-commit pre-commit run --files README.rst tests/test_pytest_rerunfailures.py changes/183.doc.rst`
- `git diff --check`


tests/test_pytest_rerunfailures.py
```diff
@@ -881,6 +881,29 @@ def test_pass_2(self, session_fixture, setup_fixture):
     assert_outcomes(result, passed=2, rerun=1)
 
 
+def test_rerun_on_package_scope_fixture_with_reruns(testdir):
```
janmrow, line 884: Just thinking about make this test to use a real package with an `__init__.py`. Because `makepyfile()` only creates a standalone test module, so this test doesn't create a real package and doesn't verify the actual package-scoped fixture lifecycle.

LouisDeconinck: Done — the test now builds a real `pkg` package with an `__init__.py` and places the test module inside it, so the package-scoped fixture binds to a real `Package` node (`pkg/test_foo.py::test_pass`). Verified: 1 passed, 1 rerun.

README.rst
```diff
@@ -72,7 +72,13 @@ maximum number of times you'd like the tests to run:
 
   $ pytest --reruns 5
 
-Failed fixture or setup_class will also be re-executed.
+Fixture support
+~~~~~~~~~~~~~~~
```
icemac: `~~~~~` is the only heading of this style in the file, so docutils makes `Fixture support` a subsection of `Re-run all failures` and everything up to the next `---` heading (the `--reruns-delay` and `--reruns-delay-backoff-factor` docs) renders nested under it. Either move this paragraph to the end of the section without its own heading, or give it a `---` heading placed after the delay docs.

— _Comment created by Claude_

README.rst
```diff
@@ -72,7 +72,13 @@ maximum number of times you'd like the tests to run:
 
   $ pytest --reruns 5
 
-Failed fixture or setup_class will also be re-executed.
+Fixture support
```
icemac: This section never reaches the documentation site: `docs/` only includes the marker-delimited fragments of the README (`SHORT-DESCRIPTION`, `INSTALLATION`, `COMPATIBILITY`, `CONTRIBUTING`), and none of the hand-written docs pages mention fixtures either. Since #183 is about users not finding this information, please wrap the section in `.. START-FIXTURES` / `.. END-FIXTURES` markers and include it from a docs page (e.g. `docs/quickstart.rst`).

— _Comment created by Claude_

tests/test_pytest_rerunfailures.py
```diff
@@ -881,6 +881,30 @@ def test_pass_2(self, session_fixture, setup_fixture):
...
```
icemac, line 1040: This test does not actually exercise package scope: the fixture lives in the single test module and is used by one test, so the identical scenario with `scope="module"` gives the same `1 passed, 1 rerun`. The `pkg/__init__.py` buys no coverage. To test what the README now documents, define the fixture in `pkg/conftest.py`, add a second test module in the package, and assert the fixture sets up twice in total and is torn down once.

— _Comment created by Claude_

changes/183.doc.rst
```diff
@@ -0,0 +1,2 @@
+Clarify support for rerunning failures in class-, module-, package-, and
```
icemac: Please keep this on a single line. `changes/README.rst` says not to hard-wrap fragments because towncrier re-wraps them, and all existing fragments are single lines.

— _Comment created by Claude_

icemac (commented):
Thanks for the PR. The new text is accurate (I verified the session-, module- and package-scope behaviour empirically), but a few things need adjusting before merge.

— _Comment created by Claude_

LouisDeconinck:
Addressed the latest review round:

- README.rst: moved the "Fixture support" paragraph after the `--reruns-delay`/`--reruns-delay-backoff-factor` docs and gave it a `---` heading, so it is a top-level section instead of a `~~~~~` subsection of "Re-run all failures".
- README.rst + docs/quickstart.rst: wrapped the section in `.. START-FIXTURES` / `.. END-FIXTURES` markers and included it from the quickstart page via `start-after`/`end-before`, matching the existing fragment includes. Verified with sphinx-build: it renders as its own section on the quickstart page.
- tests/test_pytest_rerunfailures.py: the package-scoped fixture now lives in `pkg/conftest.py` and is used by two test modules (`pkg/test_foo.py`, `pkg/test_bar.py`). The test asserts the fixture sets up twice in total (failed attempt + rerun) and is torn down once (`["setup", "setup", "teardown"]`); it fails under module scope, confirming it exercises package scope. Verified: 2 passed, 1 rerun.
- changes/183.doc.rst: unwrapped to a single line.

icemac (approved):
Thank you for this PR. 😃

PR #373, opened 2026-09-17
Include reruns in -ra/-rA short test summary

### LINKED ISSUE #45: Rerun tests are not reported with "all" setting
When I run `pytest -ra path/to/flaky/tests`, I expect rerun tests to be listed, but they aren't. I would this is because 'R' is not something that pytest know about.

This can be worked around using `pytest -raR path/to/flaky/tests`, but it would be nice if `-ra` automatically included 'R' tests as well.

LouisDeconinck:
## Summary
- `-ra`/`-rA` ("all") now shows the `rerun test summary info` section, without requiring `-rR`
- pytest expands the `a`/`A` reportchars before they reach the terminal reporter, so the plugin now checks the raw `-r` option for `a`/`A` in addition to the explicit `r`/`R`

Closes #45

#### Test plan
- [x] Added parametrized test covering `-ra` and `-rA` producing the rerun summary section
- [x] Existing `-rR` summary tests still pass


src/pytest_rerunfailures.py
```diff
@@ -1336,7 +1336,11 @@ def pytest_terminal_summary(terminalreporter):
     # Adapted from https://pytest.org/latest/_modules/_pytest/skipping.html
     tr = terminalreporter
     show_tracebacks = tr.config.getoption("rerun_show_tracebacks", False)
-    if not show_tracebacks and not any(c in "rR" for c in tr.reportchars):
+    # -rR selects reruns explicitly; -ra/-rA ("all") should include them too,
+    # but pytest expands those flags before tr.reportchars is set, so check the
+    # raw option instead.
+    requested = tr.config.option.reportchars
+    if not show_tracebacks and not any(c in "rRaA" for c in requested):
```
icemac, line 1343: Reading the raw `-r` option bypasses pytest's `N` reset, so the rerun summary now escapes a reset that used to suppress it.

`getreportopt()` treats `N` as "reset the list": for `-raN` it first expands `a` and then clears everything, leaving `reportopts == ""`. Since this check only scans the raw string for any of `rRaA`, the `a` still matches and the section is printed.

Confirmed against pytest 9.1.1 with a flaky test and `--reruns 1`:

| flags | master | this PR |
| --- | --- | --- |
| `-raN` | no section | **section printed** |
| `-ra` | no section | section printed (intended) |

Honouring only the part after the last `N` matches pytest's own semantics, and keeping `tr.reportchars` in the check preserves the explicit `R` path (plus anything that appends to `reportchars` at runtime). Full suite (252 tests) passes with this, `-raN` goes back to printing nothing, and `-rNa` still prints — consistent with pytest, which reports after a reset too.

```suggestion
    # -rR selects reruns explicitly; -ra/-rA ("all") should include them too,
    # but pytest expands those flags before tr.reportchars is set, so check the
    # raw option as well. "N" resets the list, so only honour what follows the
    # last reset.
    requested = tr.reportchars + tr.config.option.reportchars.rsplit("N", 1)[-1]
    if not show_tracebacks and not any(c in "rRaA" for c in requested):
```

— _Comment created by Claude_

PR #374, opened 2026-09-17
Disable reruns with a warning instead of erroring under --pdb

### LINKED ISSUE #248: [RFC] `--pdb`: Friendlier default (use warning and infer `--reruns 0`)
Assume `pyproject.toml`:

```toml
[tool.pytest.ini_options]
addopts = "--reruns 2"
```

# Current behavior

Right now, passing `--pdb` will raise:

```
ERROR: --reruns incompatible with --pdb
```

# Better: Assume `--reruns 0` if `--pdb`

It's highly unlikely a person attempting to drop into a shell wants reruns. Rather than error, show a warning:

`WARNING: --reruns incompatible with --pdb, --reruns will behave as if set to 0`

LouisDeconinck:
## Summary
- `--pdb` combined with `--reruns`, `--force-reruns` or the `flaky` marker now emits a `PytestWarning` ("--reruns incompatible with --pdb: reruns are disabled") instead of raising `UsageError`
- Global/ini reruns warn once at config time via `issue_config_time_warning`; marker-requested reruns warn once via `item.warn`
- `get_reruns_count` returns 0 under `--pdb`, so each test runs exactly once

Closes #248

#### Test plan
- [x] Updated the four `--pdb` tests to assert warning + single execution (using a stub `--pdbcls` so no interactive session starts)
- [x] Existing tests pass


icemac:
@LouisDeconinck Thank you for working on so many open issues. Just to inform you: I have a limited review capacity of up to one PR per day. I will review all of them but probably slower than probably expected.

— written by myself, no AI used

src/pytest_rerunfailures.py
```diff
@@ -245,9 +245,30 @@ def _get_marker(item):
...
+    in each case.
+    """
+    if getattr(config, "_pdb_reruns_warning_issued", False):
+        return
+    config._pdb_reruns_warning_issued = True
+    warning = pytest.PytestWarning(
+        "--reruns incompatible with --pdb: reruns are disabled"
+    )
+    if item is not None:
+        item.warn(warning)
```
icemac: **Warnings-as-errors turns the new warning into an `INTERNALERROR`.**

With `filterwarnings = error` in the ini (or `-W error`) the `PytestWarning` is escalated to an exception, and both emission paths abort the whole run instead of disabling reruns. Reproduced locally with a stub `--pdbcls`:

* config path (`check_options`): `pytest --reruns 1 --pdb` → the exception escapes `pytest_configure` → `INTERNALERROR`, no tests run at all.
* marker path (this line): `pytest --pdb` with `@pytest.mark.flaky(reruns=1)` → the exception escapes `pytest_runtest_protocol` → `INTERNALERROR ... pytest.PytestWarning: --reruns incompatible with --pdb: reruns are disabled`, followed by "no tests ran".

Before this PR those users got a clean `UsageError`; now they get a traceback and zero results, i.e. the feature fails exactly for projects with strict warning filters. Consider emitting the warning under `warnings.catch_warnings()` with `simplefilter("always", PytestWarning)`, or catching the escalated warning and disabling reruns anyway.

— _Comment created by Claude_

src/pytest_rerunfailures.py
```diff
@@ -245,9 +245,30 @@ def _get_marker(item):
...
+    else:
+        config.issue_config_time_warning(warning, stacklevel=2)
```
icemac: `stacklevel=2` is the right value when `issue_config_time_warning` is called directly from the deciding function, but this helper adds a frame, so the warning ends up attributed to this very line. The user-visible warnings summary is:

```
src/pytest_rerunfailures.py:264: PytestWarning: --reruns incompatible with --pdb: reruns are disabled
  config.issue_config_time_warning(warning, stacklevel=2)
```

The quoted source line is plugin internals rather than anything actionable. `stacklevel=3` would point at `check_options()` instead.

— _Comment created by Claude_

changes/248.feature.rst
```diff
@@ -0,0 +1,3 @@
+Emit a warning and disable reruns instead of raising a ``UsageError`` when
```
icemac, line 1: Replacing a hard `UsageError` (exit code 4) with a warning and a continuing run is a behaviour change users can depend on: a CI job that today fails loudly on `--reruns ... --pdb` will from now on run silently without reruns. towncrier is configured here with a `breaking` category, so `changes/248.breaking.rst` looks like the better fit (or at least a note in the feature fragment that the previous `UsageError` is gone).

— _Comment created by Claude_

icemac: No, let's keep it a feature.

src/pytest_rerunfailures.py
LouisDeconinck: Fixed in 3c0ae79 — an escalated `PytestWarning` is now caught and re-emitted under `simplefilter("always")`, then recorded through `pytest_warning_recorded`, so `-W error` / `filterwarnings = error` disables reruns cleanly instead of INTERNALERROR-ing the run. Added a regression test covering `filterwarnings = error`.

src/pytest_rerunfailures.py
LouisDeconinck: Fixed in 3c0ae79 — `stacklevel` is now 3, so the warning points at `check_options()` instead of plugin internals.

icemac (approved):
LGTM.

PR #375, opened 2026-09-17
Add --rerun-warning flag to emit a warning on each scheduled rerun

### LINKED ISSUE #318: Add ability to raise runtime warnings on reruns
I want to make the reruns more visible on CI by making use of github actions annotations.

This can be achieved easily using a plugin like [pytest-github-actions-annotate-failures/](https://pypi.org/project/pytest-github-actions-annotate-failures/) which can also translated runtime warnings into github annotations.

Now the only thing we need is to be able to configure pytest-rerunfailures to raise runtime warnings when it does reruns, so they will be visible.

AFAIK there is no option to hook into how reruns are made to make this happen.

LouisDeconinck:
## Summary
- Adds an opt-in `--rerun-warning` flag that emits a `PytestWarning` each time a test is scheduled for rerun
- Intended for CI visibility, e.g. turning warnings into annotations via pytest-github-actions-annotate-failures
- Opt-in so `filterwarnings = error` users are unaffected; the caveat is documented in `--help` and the README

Closes #318

#### Test plan
- [x] Test asserting the warning appears in the warnings summary and the rerun still happens
- [x] Test asserting no warning is emitted without the flag


src/pytest_rerunfailures.py
```diff
@@ -1337,6 +1346,13 @@ def pytest_runtest_protocol(item, nextitem):
                         continue
 
                 report.outcome = "rerun"
+                if item.session.config.option.rerun_warning:
+                    item.warn(
```
icemac, line 1350: — _Comment created by Claude_

**Bug (high): `--rerun-warning` combined with `filterwarnings = error` crashes the whole session with INTERNALERROR.**

`item.warn()` is called from `pytest_runtest_protocol`, not from inside a setup/call/teardown phase. Under `-W error` or `filterwarnings = error`, the warning is raised as an exception, and nothing catches it at this level. I reproduced it with `--reruns 1 --rerun-warning -W error` on a file where one test fails once: the run ends with `INTERNALERROR> pytest.PytestWarning: test_x.py::test_a failed on attempt 1 and will be rerun` and `no tests ran`. The remaining tests never run. The report is also left with `outcome = "rerun"` but is never logged, and any `--max-suite-reruns` slot has already been used up.

The help text and README say only that the warnings "turn into errors". Users will read that as a failing test, not an aborted session. Fix options: emit the warning in a way that cannot abort the protocol (for example, catch the exception and fail the test), or document the real behaviour clearly. A test covering `-W error` would also help.

src/pytest_rerunfailures.py
```diff
@@ -1337,6 +1346,13 @@ def pytest_runtest_protocol(item, nextitem):
                         continue
 
                 report.outcome = "rerun"
+                if item.session.config.option.rerun_warning:
```
icemac, line 1349: — _Comment created by Claude_

**Gap (low): worker-crash reruns under xdist emit no warning.**

The warning is only emitted on this in-process rerun path. `XDistHooks.pytest_handlecrashitem` also schedules reruns (`sched.mark_test_pending(crashitem)` / `report.outcome = "rerun"`) and does not check `rerun_warning`. So with `-n` and `--rerun-warning`, a test that crashes its worker and is rerun gets no annotation. The help text promises a warning "each time a test is scheduled for rerun". Either emit a warning on that path too, or narrow the wording.

PR #376, opened 2026-09-17
Document and test flaky(reruns=0) opting a test out of reruns

### LINKED ISSUE #187: Decorator to disable a test rerunning
Hi, thanks for such a helpful library! While using this library I encountered a test suit that deals badly with the rerunnig. Basically, that test spawns multiple threads on reruns and `pytest` never exits, but I want to keep the other test to be able to rerun with the provided global flags options `--rerun`, etc, so my question is is it currently possible to decorate one test suit so it wouldn't rerun on failure?

LouisDeconinck:
## Summary
- `@pytest.mark.flaky(reruns=0)` already disables reruns for a single test because the marker takes priority over the global `--reruns`/ini setting — this PR pins that behavior with regression tests and documents it
- Caveats covered: `--force-reruns` still overrides the marker, and `--reruns-mode=append` adds the marker count to the global count (so `reruns=0` does not opt out in append mode)

Closes #187

#### Test plan
- [x] `flaky(reruns=0)` + `--reruns 5` → test runs once, fails
- [x] `flaky(reruns=0)` + `--reruns-mode append` → still reruns (documented caveat)
- [x] `flaky(reruns=0)` + `--force-reruns 1` → still reruns (documented caveat)


icemac (approved):
LGTM.

Thank you for this PR. 😃

PR #377, opened 2026-09-17
Add --reruns-on-exitfirst to override rerun count under -x/--maxfail

### LINKED ISSUE #249: [RFC] `--reruns-on-exitfirst`: set reruns with `--exitfirst` / `-x`
Assume `pyproject.toml`:

```toml
[tool.pytest.ini_options]
addopts = "--reruns 2"
```

# Current behavior

Right now, passing `--exitfirst` / `-x` will rerun twice.

# Better: `--reruns-on-exitfirst 0`

This way, the user can set the reruns with `exitfirst` is used so it instantly errors. It preserves the default behavior.

`--exitfirst`'s docs:

```
  -x, --exitfirst       Exit instantly on first error or failed test
```

LouisDeconinck:
## Summary
- Adds `--reruns-on-exitfirst N`: when `-x`/`--exitfirst` or `--maxfail` is given, the rerun count for every test becomes `N` instead of the global/marker setting
- `--reruns-on-exitfirst 0` makes `-x` exit on the first failure immediately, without waiting for reruns
- `--force-reruns` still wins; negative values are rejected with a `UsageError`

Closes #249

#### Test plan
- [x] `-x --reruns-on-exitfirst 0` disables global and marker reruns
- [x] Non-zero value limits reruns under `-x`
- [x] Applies under `--maxfail` too; behavior unchanged without the option; negative value rejected


sylvesterkaczmarek (commented):
`--reruns-on-exitfirst` can enable reruns under `--pdb` without tripping the existing incompatibility check. `check_options()` only considers `--force-reruns` / the global rerun setting, while `get_reruns_count()` now returns this new value whenever `maxfail` is active. Thus `pytest -x --pdb --reruns-on-exitfirst 1` is accepted and will rerun failures even though the plugin otherwise rejects reruns with PDB. Please include the new option in the `--pdb` validation and add that no-global-reruns case to the tests.

LouisDeconinck:
Fixed in 7da4ffb: `check_options()` now treats `--reruns-on-exitfirst` as enabling reruns under `-x/--maxfail` when no other rerun count is set, so `pytest -x --pdb --reruns-on-exitfirst 1` is rejected. Added regression tests for the error and the `0` opt-out case.

LouisDeconinck:
Fixed — `check_options()` now rejects `--reruns-on-exitfirst` under `--pdb` alongside the other rerun options, with tests covering both the rejected case and the zero-value opt-out (7da4ffb).

sylvesterkaczmarek (approved):

PR #378, opened 2026-09-17
Add --xfail-flaky to report exhausted flaky failures as xfail

### LINKED ISSUE #92: Can we add functionality similar to pytest-ignore-flaky plugin
https://github.com/schettino72/pytest-ignore-flaky

Basically retry failed tests and eventually allow option to mark them as xfail instead of fail.

LouisDeconinck:
## Summary
- Adds an opt-in `--xfail-flaky` flag: a test that still fails after all its reruns are exhausted is reported as `xfailed` instead of `failed` (functionality similar to pytest-ignore-flaky, as requested)
- Only applies to tests that were actually rerun (`execution_count > 1`); tests failing on first attempt or excluded by `--only-rerun`/`--rerun-except` still report as failed
- Implemented by converting the final `call` report to `skipped` + `wasxfail`, so pytest's standard xfail reporting applies

Closes #92

#### Test plan
- [x] Always-failing test with `--reruns 2 --xfail-flaky` → 1 xfailed, 2 rerun, exit code 0
- [x] Eventual pass unaffected; tests without reruns unaffected; marker-based reruns covered
- [x] `--only-rerun` non-matching error still reports failed


sylvesterkaczmarek (commented):
`--xfail-flaky` only converts the final report when `report.when == 'call'`. Reruns can also be triggered by setup or teardown failures, so a fixture that fails on every attempt can exhaust its reruns and still finish as a normal failure even though the option says exhausted flaky failures are reported as xfailed. Could the terminal setup/teardown failure be handled too, or the option be explicitly narrowed, with a setup-failure regression test?

LouisDeconinck:
Fixed in 53b61ec: the xfail conversion now applies to the terminal failed report in any phase, so a fixture that fails on every attempt ends as xfailed. Added a setup-failure regression test.

LouisDeconinck:
Fixed — exhausted setup/teardown failures are now converted to xfail too, not just `call` reports, with a setup-failure regression test (53b61ec).

sylvesterkaczmarek (approved):

PR #380, opened 2026-09-17
Record retried failures as flakyFailure elements in JUnit XML

### LINKED ISSUE #170: Failure information for prior failing test executions is omitted from XML reports
Given a flaky test such as:

```
def test_random():
    assert bool(random.getrandbits(1))
```

The output of the JUnit XML contains no `<failure>` information when a test case fails and then subsequently passes. 

It looks like this:

```
<?xml version="1.0" encoding="utf-8"?>
<testsuites>
<testsuite name="pytest" errors="0" failures="0" skipped="0" tests="2" time="0.040" timestamp="2021-10-12T17:31:35.919038" hostname="sans">
  <testcase classname="test_sample" name="test_random" file="test_sample.py" line="6" time="0.000" />
  <testcase classname="test_sample" name="test_random" file="test_sample.py" line="6" time="0.000" />
</testsuite>
</testsuites>
```

Other test frameworks and runners typically include this failure information like this so that automated tools can process flaky test information like this in a more structured format than plain-text logs.

```
<failure message="assert False&#10; +  where False = bool(0)&#10; +    where 0 = &lt;built-in method getrandbits of Random object at 0x7f93c1078610&gt;(1)&#10; +      where &lt;built-in method getrandbits of Random object at 0x7f93c1078610&gt; = random.getrandbits">
        def test_random():
    &gt;       assert bool(random.getrandbits(1))
    E       assert False
    E        +  where False = bool(0)
    E        +    where 0 = &lt;built-in method getrandbits of Random object at 0x7f93c1078610&gt;(1)
    E        +      where &lt;built-in method getrandbits of Random object at 0x7f93c1078610&gt; = random.getrandbits

    test_sample.py:9: AssertionError</failure>
```

In order to avoid parsers mistaking the `<failure>` for deterministic, non-flaky failure, some test runners use `<flakyFailure>`. Even better, some test frameworks go so far as to mark the `<testcase>` as `flaky="true"`. However, I'm not sure this library has that level of control to achieve these things, but if it does all the better for devs trying to fix flaky tests.

LouisDeconinck:
## Summary
- pytest's junitxml plugin ignores reports with outcome `rerun`, so failures from retried attempts left no trace in `--junitxml` output
- A small per-config hook now stashes rerun reports and appends `<flakyFailure>` elements (and `<flakyError>` for setup/teardown failures) — the Maven Surefire / `flaky` plugin convention — to the test's **final** `<testcase>` when its last `call` report arrives
- No-op when `--junitxml` isn't used; works with xdist-forwarded reports since it keys on `nodeid`/`worker_id`/`item_index`/`node`

Closes #170

#### Test plan
- [x] Test that eventually passes: final testcase has 2 `flakyFailure` elements with message + traceback, no `<failure>`
- [x] Test that keeps failing: final testcase has `flakyFailure`(s) plus the real `<failure>`
- [x] Full suite passes (252 tests)


sylvesterkaczmarek (commented):
I think setup failures can still disappear here. `pending` is only flushed when a later `call` report arrives, but if the final attempt also fails in setup there is no call report, so the earlier rerun setup errors never become `<flakyError>`. Could we flush on the terminal setup/teardown report too and add that case?

LouisDeconinck:
Fixed in 8734762: `pending` is now flushed on any failed non-rerun setup/teardown report as well, so rerun setup errors become `<flakyError>` elements even when the final attempt fails in setup. Added a regression test.

LouisDeconinck:
Fixed — pending rerun reports are flushed on the terminal setup/teardown report as well, so setup errors on earlier attempts always become `<flakyError>` elements, plus a regression test (8734762).

sylvesterkaczmarek (approved):

PR #381, opened 2026-09-17
Add --max-module-reruns and --max-class-reruns scope caps

### LINKED ISSUE #213: Limit reruns per scope (module, class)
I have a test suite where all tests are somewhat flaky. I. e. 1-2 tests out of ~4000 fail almost every time. So I'd like to account for exactly that - rerun up to 2 tests, then fail.

How could the API look like? `pytest.mark.flaky` on a class/module just marks all tests as flaky - and should likely continue working this way.

LouisDeconinck:
## Summary
- Adds `--max-module-reruns N` and `--max-class-reruns N`, capping the total number of reruns within each module / test class — extending the existing suite-wide `--max-suite-reruns` mechanism to the scopes requested in the issue
- Scope counters live in `StatusDB` (in-memory) and are shared through the socket `StatusDB` under pytest-xdist, so caps hold for parallel runs; slots are reserved atomically alongside the suite slot and released if a later scope is exhausted
- The teardown-time finalizer suspension check now uses the same combined cap check, so scope exhaustion doesn't leak suspended finalizers

Closes #213

#### Test plan
- [x] Per-module cap limits reruns in one module but not others
- [x] Per-class cap limits reruns in one class but not others; function tests unaffected by `--max-class-reruns`
- [x] Negative values rejected; existing `--max-suite-reruns` tests still pass; full suite green


sylvesterkaczmarek (commented):
The class-scope key is derived with `item.nodeid.rpartition('::')[0]`, but parametrized IDs may themselves contain `::`. For example, `test.py::TestC::test_x[a::b]` produces a key ending in `test_x[a`, rather than the class node ID, so different parameter IDs can get separate counters and exceed `--max-class-reruns`. Could this derive the scope from the structured parent/class node instead of splitting the nodeid string, and add a parameter ID containing `::` as a regression?

LouisDeconinck:
Fixed in 2866443: module/class scope keys are now derived from `item.listchain()` (`pytest.Module`/`pytest.Class` nodeids) instead of splitting the nodeid, so parametrized IDs containing `::` share one counter. Added a regression test with `a::b`-style param IDs.

LouisDeconinck:
Fixed — class/module scopes are now derived from the node hierarchy instead of splitting the nodeid, so parameter IDs containing `::` get a single counter, with a regression test for `test_x[a::b]` (2866443).

sylvesterkaczmarek (approved):

PR #383, opened 2026-09-17
Re-arm pytest-timeout's timer for each rerun attempt

### LINKED ISSUE #99: Incompatible with pytest-timeout plugin
`pytest-rerunfailures` is not compatible with the [pytest-timeout](https://github.com/pytest-dev/pytest-timeout/) plugin.

Probably b/c the same pytest hooks are used.  The first time, the timeout is observed, but when the test case re-runs, there's no timeout.

`pytest-timeout`is more widely used and likely would be easier for the fix to be in this plugin if possible..

Here's a test that fails:

```python
def test_pytest_timeout_compatibility_fails(self, testdir):
    """
    Verifies the `pytest-rerunfailures` plugin is compatible with `pytest-timeout`
    plugin when the test case is not done in the allotted time.
    """
    testdir.makepyfile(
        """
        import time
        import pytest
        @pytest.mark.timeout(timeout=1)
        @pytest.mark.flaky
        def test_times_out():
            time.sleep(2)
            assert True
        """
    )
    result = testdir.runpytest()
    outcome = result.parseoutcomes()
    assert outcome.get('rerun') == 1
    assert outcome.get('failed') == 1
```
When the test case re-runs, it ignores the timeout and passes.

LouisDeconinck:
## Summary
- pytest-timeout arms a single timer around `pytest_runtest_protocol`; once it fired on the first attempt, every rerun ran without a timeout
- This plugin now detects the whole-protocol timeout (`func_only=False`) and cancels/re-arms `pytest_timeout_cancel_timer`/`pytest_timeout_set_timer` at the start of each attempt, so each attempt gets the full configured timeout (verified: a 5s test with `--timeout 1 --reruns 1` now fails in ~2s instead of passing in ~6s)
- `timeout_func_only` already worked per attempt via `pytest_runtest_call` and is untouched

Closes #99

#### Test plan
- [x] New test asserts both attempts are timed out (bounded wall-clock + `1 failed, 1 rerun`)
- [x] pytest-timeout added to tox test deps; test skips gracefully without it
- [x] Full suite green


sylvesterkaczmarek (commented):
The per-attempt timeout is re-armed before `runtestprotocol()`, but it is not cancelled when that attempt returns. `pytest-timeout` only cancels its outer timer after this entire protocol hook finishes, while this plugin can still process reports and sleep for `reruns_delay` before the next iteration. A quick failure with `--timeout 1 --reruns 1 --reruns-delay 2` can therefore time out during the backoff instead of giving the next attempt a fresh timeout. Please cancel the per-attempt timer immediately after `runtestprotocol()` (including exceptional paths), then re-arm it at the next attempt, and add a delay-greater-than-timeout regression.

LouisDeconinck:
Fixed in ec2934c: the per-attempt timer is now cancelled in a `finally` right after `runtestprotocol()` returns, before report processing and `reruns_delay`, and re-armed on the next attempt. Added a delay-greater-than-timeout regression test.

LouisDeconinck:
Fixed — the per-attempt timer is cancelled in a `finally` right after `runtestprotocol()` returns and re-armed at the next attempt, so a `reruns_delay` longer than the timeout no longer kills the next attempt; includes a delay-greater-than-timeout regression test (ec2934c).

sylvesterkaczmarek (approved):

